PR #38, opened 2023-05-20
CI: Use Rust stable

iinuwa:
This crate compiles fine with the latest Rust stable (1.69) if you remove `#![feature(option_get_or_insert_default)]` from `libwebauthn/src/lib.rs`. That feature is not used anywhere here and can be removed.


AlfioEmanueleFresta:
Thank you @iinuwa!

Issue #31, opened 2022-10-22
FIDO2 over Hybrid Transport (aka Cloud-Assisted BLE, or caBLE)


AlfioEmanueleFresta:
Blocked:
* ~~ caBLE spec has not yet been released publicly.~~
  * If you're reading this and have access to a draft, please consider sharing a copy, figuring it out from the Chromium source is not that fun :)
  * **Updated**: See CTAP 2.2 spec: https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#sctn-hybrid

Includes:

* [x] Implement Cloud-Assisted Bluetooth Low Energy (caBLE) transport
  * [x] QR-initiated transactions
  * [x] State-assisted transactions
  * [x] iOS discovery
  * [x] Android discovery
* [x] New crate for caBLE device management 
* [x] Trait for caBLE device pairing information storage
* [x] Channel clean up

martinpaljak:
This is what the Passkeys actually work with when delegating with QR code to a mobile phone?

micolous:
FYI: https://github.com/kanidm/webauthn-rs/pull/232 now has a basic implementation in Rust, which was built from reading Chromium's source code. I wrote lots of notes in there, which should aid someone else writing an implementation of the protocol, including documenting Chromium's many quirks.

It doesn't implement all the protocol features yet (like pairing, which only works on Android), but basic registration and attestation works, and it can also simulate an authenticator.

iinuwa:
FYI, the caBLE 2.1/hybrid transport spec has been published in the CTAP 2.2 review draft: https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#sctn-hybrid

AlfioEmanueleFresta:
Kudos on the reverse engineering @micolous!

Thanks for sharing this @iinuwa. I saw the draft and started implementing this a few months ago under the `cable` branch. I got advertisements working on iOS, and part of the handshake but more work is needed for the tunnel.

If anyone is interested in collaborating on this, don't hesitate to reach out.

iinuwa:
I got a dummy caBLE client through the handshake part but didn't do any of the tunnel interactions yet, so we're about at the same place. What do you have left with the handshake?

micolous:
[...]

PR #51, opened 2025-01-05
Ref #31: Initial support for Hybrid transport (iOS only, non persistent)

AlfioEmanueleFresta:
An initial rough implementation of hybrid transport - formerly known as Cloud Assisted Bluetooth Low Energy, or caBLE.

## Includes

* Generates QR code for QR-initiated transactions
* Adds WebAuthn example which prints QR code to terminal (`examples/webauthn_cable.rs`)
* Adds submodule fork of `snow` (noise protocol framework) with NIST P-256 curve implementation, as required by CTAP spec (see https://github.com/mcginty/snow/pull/185)

## Limitations and follow ups

* Unable to discover Android devices (issue TBC)
* Does not yet support state-assisted transactions (part of #31)
* Clean up (part of #31)


AlfioEmanueleFresta:
@msirringhaus would love your thoughts if you have the time for a review. Will try and merge this by tomorrow EOD otherwise.

libwebauthn/src/proto/ctap2/cbor/response.rs
```diff
@@ -4,12 +4,21 @@ use std::convert::{TryFrom, TryInto};
...
 pub struct CborResponse {
     pub status_code: CtapError,
     pub data: Option<Vec<u8>>,
 }
 
+impl CborResponse {
+    pub fn new_success_from_slice(slice: &[u8]) -> Self {
+        Self {
+            status_code: CtapError::Ok,
+            data: Some(slice.to_vec()),
```
msirringhaus: Can it happen here that slice is empty and we want `data: None` here?

libwebauthn/src/transport/cable/crypto.rs
```diff
@@ -0,0 +1,79 @@
...
+    plaintext[0] == 0
+}
+
+#[instrument]
+pub fn trial_decrypt_advert(eid_key: &[u8], candidate_advert: &[u8]) -> Option<Vec<u8>> {
+    if candidate_advert.len() != 20 {
+        warn!("candidate advert is not 20 bytes");
+        return None;
+    }
+
```
msirringhaus, line 36: Should we also check the length of `eid_key` here? Should be 64, right?

libwebauthn/src/transport/cable/digit_encode.rs
```diff
@@ -0,0 +1,41 @@
+const CHUNK_SIZE: usize = 7;
+const CHUNK_DIGITS: usize = 17;
+const ZEROS: &str = "00000000000000000";
+
+/// The number of digits needed to encode each length of trailing data from 6 bytes down to zero,
+/// i.e. it’s 15, 13, 10, 8, 5, 3, 0 written in hex.
+const PARTIAL_CHUNK_DIGITS: usize = 0x0fda8530;
```
msirringhaus, line 7: Wouldn't an array be simpler here?

libwebauthn/src/transport/cable/digit_encode.rs
```diff
@@ -0,0 +1,41 @@
...
+        let mut chunk = [0u8; 8];
+        chunk[..CHUNK_SIZE].copy_from_slice(&input[..CHUNK_SIZE]);
+        let v = u64::from_le_bytes(chunk);
+        let v = v.to_string();
+        output.push_str(&ZEROS[..CHUNK_DIGITS - v.len()]);
+        output.push_str(&v);
+        input = &input[CHUNK_SIZE..];
+    }
+    if !input.is_empty() {
+        let digits = 15 & (PARTIAL_CHUNK_DIGITS >> (4 * input.len()));
```
msirringhaus: If we keep the long hex-number, maybe at least write 15 also in hex? Then it's a bit faster to grok (for me).

libwebauthn/src/transport/cable/qr_code_device.rs
```diff
@@ -0,0 +1,283 @@
...
+            QrCodeOperationHint::GetAssertionRequest => serializer.serialize_str("ga"),
+            QrCodeOperationHint::MakeCredential => serializer.serialize_str("mc"),
+        }
+    }
+}
+
+#[derive(Debug, SerializeIndexed)]
+pub struct CableQrCode {
+    // Key 0: a 33-byte, P-256, X9.62, compressed public key.
+    pub public_key: ByteBuf,
```
msirringhaus: Would it make sense to use `ByteArray` here, if it's a fixed length array? (Same for `qr_secret` below).

libwebauthn/examples/webauthn_cable.rs
```diff
@@ -0,0 +1,143 @@
...
+#[tokio::main]
+pub async fn main() -> Result<(), Box<dyn Error>> {
+    setup_logging();
+
+    let _device_info_store: Box<dyn CableKnownDeviceInfoStore> =
+        Box::new(EphemeralDeviceInfoStore::default());
+
+    // Create QR code
+    let mut device: CableQrCodeDevice<'_> =
+        CableQrCodeDevice::new_transient(QrCodeOperationHint::MakeCredential);
```
msirringhaus, line 43: Is `new_transient()` the same as `new()` + `EphemeralDeviceInfoStore`? If so, couldn't you simply require a store for a leaner API?

msirringhaus (commented):
Some minor things. I didn't do a full review if everything is according to the spec, though. I'm guessing, if it works, it can't be that wrong :)

libwebauthn/src/transport/cable/digit_encode.rs
AlfioEmanueleFresta: I agree, this code comes directly from the specs ([11.5.1. QR-initiated Transactions](https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#hybrid-qr-initiated)), but it can certainly be improved.

libwebauthn/examples/webauthn_cable.rs
AlfioEmanueleFresta: I see your point. IMO, keeping EphemeralDeviceInfoStore an internal implementation detail makes the API slightly simpler for clients, as they would either not care about persistence (in which case they would use `new_transient()`), or want to implement their own (`new(...)` with a concrete impementation).

AlfioEmanueleFresta:
Thanks for your feedback @msirringhaus! Addressed your comments - I'll rebase and land tomorrow.

Issue #53, opened 2025-01-05
Project update (draft)


AlfioEmanueleFresta:
@iinuwa is leading the work on the portal API definition and implementation, as part of [linux-webauthn-platform-api](https://github.com/iinuwa/linux-webauthn-platform-api). 

This project will narrow its focus on providing WebAuthn and CTAP APIs, and Linux transport implementations, to support a platform authenticator akin to Windows Hello. It will be renamed accordingly.

More details to follow.

Includes:

* [ ] README.md changes
  * [ ] Describe overall project and goals
  * [ ] Link to @iinuwa's portal
  * [ ] Comparison to other libraries and key features (eg hybrid transport for passkeys)
* [ ] Change project name (libwebauthn-linux?)
* [ ] Move project to org?

PR #55, opened 2025-01-14
Implement AuthenticatorConfig

msirringhaus:
Refactored `user_verification()` to split out the calculation of `uv_auth_param` so that other commands can implement their own algorithms for it in `calculate_and_set_uv_auth()`.

Needs probably a rebase, once (if) #48 gets merged.

### DIFF
--- a/libwebauthn/src/management.rs
+++ b/libwebauthn/src/management.rs
@@ -1,0 +1,198 @@
+use serde_cbor::ser::to_vec;
+use std::time::Duration;
+
+use crate::pin::{PinProvider, PinUvAuthProtocol};
+use crate::proto::ctap2::Ctap2AuthenticatorConfigCommand;
+pub use crate::transport::error::{CtapError, Error, TransportError};
+use crate::transport::Channel;
+use crate::webauthn::user_verification;
+use crate::{
+    ops::webauthn::UserVerificationRequirement,
+    proto::ctap2::{
+        ClientPinRequestPermissions, Ctap2, Ctap2AuthenticatorConfigRequest,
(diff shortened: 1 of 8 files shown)


AlfioEmanueleFresta:
Thank you @msirringhaus, I will review this by tomorrow EOD.

libwebauthn/src/management.rs
```diff
@@ -0,0 +1,198 @@
...
+        ClientPinRequestPermissions, Ctap2, Ctap2AuthenticatorConfigRequest,
+        Ctap2UserVerifiableRequest,
+    },
+};
+use async_trait::async_trait;
+use serde_bytes::ByteBuf;
+
+#[async_trait]
+pub trait AuthenticatorConfig {
+    async fn toogle_always_uv(
```
AlfioEmanueleFresta, line 21: Typo: toggle_always_uv

AlfioEmanueleFresta: Typo: toggle_always_uv

AlfioEmanueleFresta (approved):

libwebauthn/src/management.rs
msirringhaus: Whoops.

AlfioEmanueleFresta:
Thank you for your work @msirringhaus! Merging.

We should add PIN management + authenticator config functionality to the README. We could do this as part of #53.

msirringhaus:
I was thinking about also adding all the other ctap-stuff like BioEnrollment, CredManagement, etc. as well. Or would you rather want me to also play with the caBLE-functionality?

PR #56, opened 2025-01-21
Catch UvBlocked and have way to fall back to PIN internally

msirringhaus:
Currently, there is no way to recover from a `UvBlocked`-error.
I added an internal fallback to PIN, if we encounter an `UvBlocked`-error.

Alternatively, we could also mark that error as retryable, return it to the application as normal and give the API some way to communicate back to skip Uv and use PIN instead, if the app decides to retry. But this seemed to me a rather clunky API, as I can't think of a scenario, where one wouldn't want to fall back to PIN.


libwebauthn/src/webauthn.rs
```diff
@@ -19,6 +19,7 @@ use crate::proto::ctap2::{
     Ctap2, Ctap2ClientPinRequest, Ctap2GetAssertionRequest, Ctap2GetInfoResponse,
     Ctap2MakeCredentialRequest, Ctap2UserVerifiableRequest, Ctap2UserVerificationOperation,
 };
+use crate::transport::error::PlatformError;
```
AlfioEmanueleFresta: Let's re-export this as per `CtapError`, `Error` and `TransportError` below.

libwebauthn/src/webauthn.rs
```diff
@@ -248,60 +249,84 @@ where
...
+            }
+        };
+
+        match channel.ctap2_client_pin(&token_request, timeout).await {
+            Ok(t) => {
+                break (uv_proto, t, shared_secret);
+            }
+            // Internal retry, because we otherwise can't fall back to PIN, if the UV is blocked
+            Err(Error::Ctap(CtapError::UvBlocked)) => {
+                uv_blocked = true;
```
AlfioEmanueleFresta, line 321: It may be worth adding a `warn!` or `info!` here

libwebauthn/src/proto/ctap2/model.rs
```diff
@@ -644,26 +644,32 @@ impl Ctap2GetInfoResponse {
...
+                );
+            } else if self.option_enabled("clientPin") {
                 // !pinUvAuthToken
-                assert!(self.option_enabled("clientPin"));
                 debug!("getPinToken");
-                return Ctap2UserVerificationOperation::GetPinToken;
+                return Some(Ctap2UserVerificationOperation::GetPinToken);
+            } else {
+                // No UV and no PIN (e.g. maybe UV was blocked and no PIN available)
+                return None;
```
AlfioEmanueleFresta, line 672: Could you add a `debug!` statement?

libwebauthn/src/proto/ctap2/model.rs
```diff
@@ -644,26 +644,32 @@ impl Ctap2GetInfoResponse {
             (self.option_enabled("pinUvAuthToken") && self.option_enabled("uv"))
     }
 
-    pub fn uv_operation(&self) -> Ctap2UserVerificationOperation {
-        if self.option_enabled("uv") {
+    pub fn uv_operation(&self, uv_blocked: bool) -> Option<Ctap2UserVerificationOperation> {
```
AlfioEmanueleFresta, line 647: This method is getting a bit complex - may be worth refactoring this soon (as part of this PR or as a new issue)

AlfioEmanueleFresta (approved):
Thank you for these changes @msirringhaus!

I agree with the overall approach. 

Have you tested this with any devices that support both built-in UV and PIN? If so, which one?

Minor comments, mostly to sprinkle a few log statements. 

I'm unsure if and how built-in UV is handled for hybrid transport passkeys on phones. Once the hybrid transport changes are complete, I think it will be worth testing. In the unlikely case the PIN-retry causes an additional prompt to the user, it may be worth considering changing this behaviour - or making this configurable.

libwebauthn/src/webauthn.rs
msirringhaus: Yes, makes sense!

libwebauthn/src/webauthn.rs
msirringhaus: True

libwebauthn/src/proto/ctap2/model.rs
msirringhaus: I converted the comment to a debug-statement

libwebauthn/src/proto/ctap2/model.rs
msirringhaus: Let's make an issue for this

msirringhaus:
Thanks for the review!

I have tested it with a Yubikey Bio.

With respect to hybrid mode: I'm also not yet sure how this will be affected.

libwebauthn/src/proto/ctap2/model.rs
AlfioEmanueleFresta: Created #60

AlfioEmanueleFresta:
Thank you for your changes and for addressing the feedback!

PR #57, opened 2025-01-24
Add possibility to cache pinUvAuthToken in PinProvider

msirringhaus:
This is mostly prep-work for upcoming features like BioEnrollments, where we would want to avoid having to enter the PIN for each fingerprint-sample, when enrolling a new fingerprint. Although the `webauthn_hid`-example can already skip one PIN-entry now, too.

I'm aware that `PinProvider` might not be the greatest name now for this trait, but I think it's the right place to add the caching, since it is very related to PINs.

Note: Needs rebase, once #56 lands.

### DIFF
--- a/libwebauthn/src/proto/ctap2/model.rs
+++ b/libwebauthn/src/proto/ctap2/model.rs
@@ -881,6 +881,7 @@ impl Ctap2ClientPinRequest {
 }
 
 bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
     pub struct ClientPinRequestPermissions: u32 {
         const MAKE_CREDENTIAL = 0x01;
         const GET_ASSERTION = 0x02;
@@ -892,7 +893,7 @@ bitflags! {
 }
 
 #[repr(u32)]
-#[derive(Debug, Clone, Copy, FromPrimitive, PartialEq, Serialize_repr, Deserialize_repr)]
+#[derive(Debug, Clone, Copy, FromPrimitive, PartialEq, Eq, Serialize_repr, Deserialize_repr)]
 pub enum Ctap2PinUvAuthProtocol {
     One = 1,
     Two = 2,
--- a/libwebauthn/src/proto/ctap2/model.rs
+++ b/libwebauthn/src/proto/ctap2/model.rs
@@ -478,7 +478,7 @@ pub trait Ctap2UserVerifiableRequest {
         uv_auth_token: &[u8],
     );
     fn client_data_hash(&self) -> &[u8];
-    fn permissions(&self) -> ClientPinRequestPermissions;
+    fn permissions(&self) -> Ctap2AuthTokenPermissionRole;
     fn permissions_rpid(&self) -> Option<&str>;
 }
 
@@ -506,10 +506,10 @@ impl Ctap2UserVerifiableRequest for Ctap2MakeCredentialRequest {
         self.hash.as_slice()
     }
 
-    fn permissions(&self) -> ClientPinRequestPermissions {
+    fn permissions(&self) -> Ctap2AuthTokenPermissionRole {
         // GET_ASSERTION needed for pre-flight requests
-        return ClientPinRequestPermissions::MAKE_CREDENTIAL
-            | ClientPinRequestPermissions::GET_ASSERTION;
+        return Ctap2AuthTokenPermissionRole::MAKE_CREDENTIAL
+            | Ctap2AuthTokenPermissionRole::GET_ASSERTION;
     }
 
     fn permissions_rpid(&self) -> Option<&str> {
@@ -539,8 +539,8 @@ impl Ctap2UserVerifiableRequest for Ctap2GetAssertionRequest {
         self.client_data_hash.as_slice()
     }
 
-    fn permissions(&self) -> ClientPinRequestPermissions {
-        return ClientPinRequestPermissions::GET_ASSERTION;
+    fn permissions(&self) -> Ctap2AuthTokenPermissionRole {
+        return Ctap2AuthTokenPermissionRole::GET_ASSERTION;
     }
 
     fn permissions_rpid(&self) -> Option<&str> {
@@ -801,7 +801,7 @@ impl Ctap2ClientPinRequest {
         protocol: Ctap2PinUvAuthProtocol,
         public_key: PublicKey,
         pin_hash_enc: &[u8],
-        permissions: ClientPinRequestPermissions,
+        permissions: Ctap2AuthTokenPermissionRole,
         permissions_rpid: Option<&str>,
     ) -> Self {
         Self {
@@ -821,7 +821,7 @@ impl Ctap2ClientPinRequest {
     pub fn new_get_uv_token_with_perm(
         protocol: Ctap2PinUvAuthProtocol,
         public_key: PublicKey,
-        permissions: ClientPinRequestPermissions,
+        permissions: Ctap2AuthTokenPermissionRole,
         permissions_rpid: Option<&str>,
     ) -> Self {
         Self {
@@ -882,7 +882,7 @@ impl Ctap2ClientPinRequest {
 
 bitflags! {
     #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-    pub struct ClientPinRequestPermissions: u32 {
+    pub struct Ctap2AuthTokenPermissionRole: u32 {
         const MAKE_CREDENTIAL = 0x01;
         const GET_ASSERTION = 0x02;
         const CREDENTIAL_MANAGEMENT = 0x04;
--- a/libwebauthn/src/transport/channel.rs
+++ b/libwebauthn/src/transport/channel.rs
@@ -1,7 +1,7 @@
 use std::fmt::{Debug, Display};
 use std::time::Duration;
 
-use crate::proto::ctap2::{ClientPinRequestPermissions, Ctap2PinUvAuthProtocol};
+use crate::proto::ctap2::{Ctap2AuthTokenPermissionRole, Ctap2PinUvAuthProtocol};
 use crate::proto::{
     ctap1::apdu::{ApduRequest, ApduResponse},
     ctap2::cbor::{CborRequest, CborResponse},
@@ -33,33 +33,33 @@ pub trait Channel: Send + Sync + Display + Ctap2AuthTokenStore {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Ctap2AuthTokenIdentifier {
+pub struct Ctap2AuthTokenPermission {
     pin_uv_auth_protocol: Ctap2PinUvAuthProtocol,
-    permissions: ClientPinRequestPermissions,
-    permissions_rpid: Option<String>,
+    role: Ctap2AuthTokenPermissionRole,
+    rpid: Option<String>,
 }
 
-impl Ctap2AuthTokenIdentifier {
+impl Ctap2AuthTokenPermission {
     pub fn new(
         pin_uv_auth_protocol: Ctap2PinUvAuthProtocol,
-        permissions: ClientPinRequestPermissions,
+        permissions: Ctap2AuthTokenPermissionRole,
         permissions_rpid: Option<&str>,
     ) -> Self {
         Self {
             pin_uv_auth_protocol,
-            permissions,
-            permissions_rpid: permissions_rpid.map(str::to_string),
+            role: permissions,
+            rpid: permissions_rpid.map(str::to_string),
         }
     }
 
-    pub fn is_satisfied_by(&self, other: &Ctap2AuthTokenIdentifier) -> bool {
-        if self.pin_uv_auth_protocol != other.pin_uv_auth_protocol {
+    pub fn contains(&self, requested: &Ctap2AuthTokenPermission) -> bool {
+        if self.pin_uv_auth_protocol != requested.pin_uv_auth_protocol {
             return false;
         }
-        if self.permissions_rpid != other.permissions_rpid {
+        if self.rpid != requested.rpid {
             return false;
         }
-        self.permissions.contains(other.permissions)
+        self.role.contains(requested.role)
     }
 }
 
@@ -67,9 +67,9 @@ impl Ctap2AuthTokenIdentifier {
 pub trait Ctap2AuthTokenStore {
     fn store_uv_auth_token(
         &mut self,
-        identifier: Ctap2AuthTokenIdentifier,
+        permission: Ctap2AuthTokenPermission,
         pin_uv_auth_token: &[u8],
     );
-    fn get_uv_auth_token(&self, identifier: &Ctap2AuthTokenIdentifier) -> Option<&[u8]>;
+    fn get_uv_auth_token(&self, requested_permission: &Ctap2AuthTokenPermission) -> Option<&[u8]>;
     fn clear_uv_auth_token_store(&mut self);
 }
--- a/libwebauthn/src/transport/channel.rs
+++ b/libwebauthn/src/transport/channel.rs
@@ -1,6 +1,7 @@
 use std::fmt::{Debug, Display};
 use std::time::Duration;
 
+use crate::proto::ctap2::{ClientPinRequestPermissions, Ctap2PinUvAuthProtocol};
 use crate::proto::{
     ctap1::apdu::{ApduRequest, ApduResponse},
     ctap2::cbor::{CborRequest, CborResponse},
@@ -19,7 +20,7 @@ pub enum ChannelStatus {
 }
 
 #[async_trait]
-pub trait Channel: Send + Sync + Display {
+pub trait Channel: Send + Sync + Display + Ctap2AuthTokenStore {
     async fn supported_protocols(&self) -> Result<SupportedProtocols, Error>;
     async fn status(&self) -> ChannelStatus;
     async fn close(&mut self);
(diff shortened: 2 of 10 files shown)


AlfioEmanueleFresta:
Thanks for this PR @msirringhaus.

From my reading of the specs, it looks like the platform is responsible for keeping track of these tokens and, most importantly, ensuring they are scoped correctly to both operations, and RPIDs.

> [6.5.5.7. Operations to Obtain a pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#gettingPinUvAuthToken)
> [...] Once the platform obtains a [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken), it can be used in subsequent authenticator operations for the length of its [max usage time period](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-max-usage-time-period) (see [§ 6.5.2.1 pinUvAuthToken State](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#authnrClientPin-globalState-puat)), thereby avoiding asking the user for verification for each authenticator operation.
> [...] When obtaining a [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken), the platform requests [permissions](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#perms-param-permissions) appropriate for the operations it intends to perform. **Consequently, the [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken) can only be used for those operations.** Some permissions require the presence of the RPID parameter, known as a permissions RPID. See also [§ 6.5.2.1 pinUvAuthToken State](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#authnrClientPin-globalState-puat).

Furthermore, this note in [6.5. authenticatorClientPIN (0x06)](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#authenticatorClientPIN) suggests authenticator can expire tokens arbitrarily, requiring a retry after a new token is obtained:

> Note: The platform may flexibly manage the lifetime of its copy of the [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken) based on the usage scenario. However, it should erase its copy of the pinUvAuthToken as soon as possible when it is no longer needed. The authenticator also can expire the pinUvAuthToken based on certain conditions like changing a PIN, authenticator timeouts, the platform system waking up from a suspend state, etc. If the pinUvAuthToken has expired, the authenticator will return CTAP2_ERR_PIN_TOKEN_EXPIRED and the platform can act on the error accordingly, e.g., by [getting a new pinUvAuthToken from the authenticator](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#gettingPinUvAuthToken).

Both token management and retries are described as platform behaviours in CTAP, so I believe we should take this responsibility within the library - rather than delegating it to clients. By contrast, `PinProvider` is meant to delegate PIN entering to clients, so I don't think it's appropriate.

We could introduce an additional trait for clients to provide persistence (similarly to how the client will need to store 'known authenticators' for hybrid transport). However, I'm not convinced it's necessary at this stage.

The docs seem to suggest the tokens are short lived. In this codebase, *channels* represents a session, or a connection, to an authenticator - and seems to fit the use case best. We could extend the channel trait, or possibly introduce a token manager trait and implement this for all channels.

In summary, I propose:
* Adding a new token management trait
* Implementing the trait ourselves, hoisting state to the transport channel
* Scoping tokens by permission and RPIDs
* Retrying requests automatically if the authenticator reports a token expires

WDYT?

AlfioEmanueleFresta (changes requested):
Thanks for this PR @msirringhaus.

From my reading of the specs, it looks like the platform is responsible for keeping track of these tokens and, most importantly, ensuring they are scoped correctly to both operations, and RPIDs.

> [6.5.5.7. Operations to Obtain a pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#gettingPinUvAuthToken)
> [...] Once the platform obtains a [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken), it can be used in subsequent authenticator operations for the length of its [max usage time period](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-max-usage-time-period) (see [§ 6.5.2.1 pinUvAuthToken State](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#authnrClientPin-globalState-puat)), thereby avoiding asking the user for verification for each authenticator operation.
> [...] When obtaining a [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken), the platform requests [permissions](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#perms-param-permissions) appropriate for the operations it intends to perform. **Consequently, the [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken) can only be used for those operations.** Some permissions require the presence of the RPID parameter, known as a permissions RPID. See also [§ 6.5.2.1 pinUvAuthToken State](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#authnrClientPin-globalState-puat).

Furthermore, this note in [6.5. authenticatorClientPIN (0x06)](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#authenticatorClientPIN) suggests authenticator can expire tokens arbitrarily, requiring a retry after a new token is obtained:

> Note: The platform may flexibly manage the lifetime of its copy of the [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken) based on the usage scenario. However, it should erase its copy of the pinUvAuthToken as soon as possible when it is no longer needed. The authenticator also can expire the pinUvAuthToken based on certain conditions like changing a PIN, authenticator timeouts, the platform system waking up from a suspend state, etc. If the pinUvAuthToken has expired, the authenticator will return CTAP2_ERR_PIN_TOKEN_EXPIRED and the platform can act on the error accordingly, e.g., by [getting a new pinUvAuthToken from the authenticator](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#gettingPinUvAuthToken).

Both token management and retries are described as platform behaviours in CTAP, so I believe we should take this responsibility within the library - rather than delegating it to clients. By contrast, `PinProvider` is meant to delegate PIN entering to clients, so I don't think it's appropriate.

We could introduce an additional trait for clients to provide persistence (similarly to how the client will need to store 'known authenticators' for hybrid transport). However, I'm not convinced it's necessary at this stage.

The docs seem to suggest the tokens are short lived. In this codebase, *channels* represents a session, or a connection, to an authenticator - and seems to fit the use case best. We could extend the channel trait, or possibly introduce a token manager trait and implement this for all channels.

In summary, I propose:
* Adding a new token management trait
* Implementing the trait ourselves, hoisting state to the transport channel
* Scoping tokens by permission and RPIDs
* Retrying requests automatically if the authenticator reports a token expires

WDYT?

msirringhaus:
To be honest, I was thinking about something like this in the beginning, but simply didn't know how much of the business-logic was supposed to be inside the lib and how much outside (i.e. what counts as "platform"). 
If we want this inside the lib, I'm in favor of your suggestion, and try to get this working.

msirringhaus:
Ok, I know this isn't the prettiest with the loops everywhere, but couldn't think of a cleaner way to do this, today. I'm very open to suggestions :) 

I implemented only a no-op PinUvAuthTokenStorage for Cable, since I'm not yet 100% sure how that would work there.

libwebauthn/src/transport/channel.rs
```diff
@@ -30,3 +31,45 @@ pub trait Channel: Send + Sync + Display {
     async fn cbor_send(&mut self, request: &CborRequest, timeout: Duration) -> Result<(), Error>;
     async fn cbor_recv(&mut self, timeout: Duration) -> Result<CborResponse, Error>;
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Ctap2AuthTokenIdentifier {
```
AlfioEmanueleFresta: If we wanted to match the spec terminology, we could use `Ctap2AuthTokenPermissions`, with protocol, role, and rpid (optional).

libwebauthn/src/proto/ctap2/model.rs
```diff
@@ -881,6 +881,7 @@ impl Ctap2ClientPinRequest {
 }
 
 bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
     pub struct ClientPinRequestPermissions: u32 {
```
AlfioEmanueleFresta: This is no longer just for ClientPinRequest, should we rename this 'permissions role' to match [6.5.5.7](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#gettingPinUvAuthToken)?

libwebauthn/src/transport/channel.rs
```diff
@@ -30,3 +31,45 @@ pub trait Channel: Send + Sync + Display {
...
+            permissions,
+            permissions_rpid: permissions_rpid.map(str::to_string),
+        }
+    }
+
+    pub fn is_satisfied_by(&self, other: &Ctap2AuthTokenIdentifier) -> bool {
+        if self.pin_uv_auth_protocol != other.pin_uv_auth_protocol {
+            return false;
+        }
+        if self.permissions_rpid != other.permissions_rpid {
```
AlfioEmanueleFresta: Frustratingly, the spec is quite convoluted and it would seem the behaviour changes depending on the permissions role:
* MakeCredential, GetAssertion: RPID has to be present and match.
* CredentialManagement: Here, the RPID is optional and used to restrict the permissions. Crucially, a request with a specific RPID, would be allowed by a pre-existing token _with an empty RPID_.
  > [6.5.5.7](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#gettingPinUvAuthToken): The RPID parameter is optional, if it is present, the [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#puatoken-pinuvauthtoken) can only be used for Credential Management operations on Credentials associated with that RPID.
* BioEnrollment, LargeBlobWrite, AuthenticatorConfiguration: The RPID should be ignored.

libwebauthn/src/transport/channel.rs
```diff
@@ -30,3 +31,45 @@ pub trait Channel: Send + Sync + Display {
...
+        permissions_rpid: Option<&str>,
+    ) -> Self {
+        Self {
+            pin_uv_auth_protocol,
+            permissions,
+            permissions_rpid: permissions_rpid.map(str::to_string),
+        }
+    }
+
+    pub fn is_satisfied_by(&self, other: &Ctap2AuthTokenIdentifier) -> bool {
```
AlfioEmanueleFresta: I'm not a native English speaker, but I find this terminology (identifier <- is satisfied by <- `other` identifier) a bit confusing.

I wonder if it could be simplified further, e.g.:
* identifier <- is_subset_of <- `other` identifier
* permissions -> allows -> permissions `request`

libwebauthn/src/transport/hid/channel.rs
```diff
@@ -443,3 +447,26 @@ bitflags! {
...
+        &mut self,
+        identifier: Ctap2AuthTokenIdentifier,
+        pin_uv_auth_token: &[u8],
+    ) {
+        self.auth_token_storage = Some((identifier, pin_uv_auth_token.to_vec()));
+    }
+
+    fn get_uv_auth_token(&self, identifier: &Ctap2AuthTokenIdentifier) -> Option<&[u8]> {
+        if let Some((id, token)) = &self.auth_token_storage {
+            if id.is_satisfied_by(identifier) {
```
AlfioEmanueleFresta: I would read this as meaning `id` is satisfied by existing permissions `identifier`, but I think this means the opposite.

libwebauthn/src/transport/hid/channel.rs
```diff
@@ -50,6 +52,7 @@ pub struct HidChannel<'d> {
     device: &'d HidDevice,
     open_device: OpenHidDevice,
     init: InitResponse,
+    auth_token_storage: Option<(Ctap2AuthTokenIdentifier, Vec<u8>)>,
```
AlfioEmanueleFresta: Nit: This could just be called `auth_token`.

AlfioEmanueleFresta (commented):
Thank you so much for updating this PR! 

Left a few comments, mostly naming.

I don't currently write Rust very frequently, I hope we could find a way one day to remove the macro, and possibly write the storage once generically over any channel types (rather than having to re-implement the trait). Neither are a blocker though.

libwebauthn/src/transport/channel.rs
msirringhaus: I was actually struggling with the naming here, so thanks for the suggestion :)

libwebauthn/src/proto/ctap2/model.rs
msirringhaus: Makes sense, yes

libwebauthn/src/transport/channel.rs
msirringhaus: > MakeCredential, GetAssertion: RPID has to be present and match.

Not necessarily. One can use a PinAuthToken without RPID and there is a fallback (6.1.2.11):
> 7. If the [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#puatoken-pinuvauthtoken) does not have a [permissions RP ID](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#permissions-rp-id) associated:
>
>     1\. Associate the request’s [rp.id](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#makecred-rpid) parameter value with the [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#puatoken-pinuvauthtoken) as its [permissions RP ID](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#permissions-rp-id).

So the whole thing is rather messy, and I'm wondering if the initial, stupid "just compare the RPIDs" isn't a good enough approximation. Worst case scenario is that we do not reuse a token in edge cases were we could reuse it, prompting the user for a PIN. The other potential problem is that we reuse a token that shouldn't be reused. Then the device will decline the request, and we repeat the request automatically with a fresh token. So the user wouldn't even notice it.

Most of these constellations probably won't ever happen anyways. The library code surrounding it, doesn't fully utilize all the possible permuations, e.g. I doubt, we'll ever store an auth-token that has both a BioEnrollment-role and contains an RPID, because that is not how the library is designed.

WDYT?

libwebauthn/src/transport/channel.rs
msirringhaus: Yes, I was not at all happy with my choice of words here, either. 

And, as you pointed out below, I got it backwards as well. I should have used `satisfies()` instead of `is_satisfied_by()`.

`allows()` sounds reasonable. Although, right now I'm also thinking about `contains()`:
```
if stored_permission.contains(requested_permission) {
    return stored_token;
}
```

I'd be happy with both. WDYT?

AlfioEmanueleFresta (approved):

AlfioEmanueleFresta:
CI was broken, fixed in  #62. Rebasing.

PR #65, opened 2025-02-05
Fix getRetries CTAP 2.0 / 2.1 behavior

msirringhaus:
For some reason I have not tried my Nitrokey3 with this lib before and noticed some issues.

1. The GetInfo response contained already some 2.2 members, which made serde unhappy, so I added all of the 2.2 fields (although, we _should_ ignore all unknown keys, but serde_indexed doesn't seem to have this functionality)
2. `getRetries` requires the PIN protocol in CTAP 2.0, but not anymore in 2.1.
3. Sadly, my NK3 still error-ed out with MissingParameter, because it has a bug and reports full 2.1 support, but still requires the PIN protocol (see [here](https://github.com/Nitrokey/nitrokey-fido2-firmware/issues/77)). So I changed the hard error into a soft error, as the displaying of the retry counter is optional anyways.

### DIFF
--- a/libwebauthn/src/pin.rs
+++ b/libwebauthn/src/pin.rs
@@ -463,9 +463,20 @@ where
             return Err(Error::Platform(PlatformError::PinTooLong));
         }
 
+        let uv_proto = select_uv_proto(&get_info_response).await?;
+
         let current_pin = match get_info_response.options.as_ref().unwrap().get("clientPin") {
             // Obtaining the current PIN, if one is set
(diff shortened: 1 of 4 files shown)


AlfioEmanueleFresta (approved):

AlfioEmanueleFresta:
Thanks @msirringhaus! Good point on ignoring unknown fields, IMO this is crucial to do before our first release, allowing forward-compat with future spec versions. Created #66.

PR #67, opened 2025-02-11
Implement Credential Management

### LINKED ISSUE #24: FIDO 2.1: Credential management
https://fidoalliance.org/specs/fido-v2.1-rd-20210309/fido-client-to-authenticator-protocol-v2.1-rd-20210309.html#authenticatorCredentialManagement

msirringhaus:
Fixes #24

### DIFF
--- a/libwebauthn/examples/cred_management.rs
+++ b/libwebauthn/examples/cred_management.rs
@@ -1,0 +1,237 @@
+use libwebauthn::management::CredentialManagement;
+use libwebauthn::pin::{PinProvider, StdinPromptPinProvider};
+use libwebauthn::proto::ctap2::{Ctap2, Ctap2CredentialData, Ctap2PublicKeyCredentialRpEntity};
+use libwebauthn::proto::CtapError;
+use libwebauthn::transport::hid::list_devices;
+use libwebauthn::transport::Device;
+use libwebauthn::webauthn::Error as WebAuthnError;
+use serde_bytes::ByteBuf;
+use std::fmt::Display;
+use std::io::{self, Write};
+use std::time::Duration;
+use text_io::read;
+use tracing_subscriber::{self, EnvFilter};
+
+const TIMEOUT: Duration = Duration::from_secs(10);
+
+fn setup_logging() {
+    tracing_subscriber::fmt()
+        .with_env_filter(EnvFilter::from_default_env())
+        .without_time()
+        .init();
+}
+
+macro_rules! handle_retries {
+    ($res:expr) => {
+        loop {
+            match $res.await {
+                Ok(r) => break r,
+                Err(WebAuthnError::Ctap(ctap_error)) => {
+                    if ctap_error.is_retryable_user_error() {
+                        println!("Oops, try again! Error: {}", ctap_error);
+                        continue;
+                    }
+                    return Err(WebAuthnError::Ctap(ctap_error));
+                }
+                Err(err) => return Err(err),
+            }
+        }
+    };
+}
+
+fn format_rp(rp: &Ctap2PublicKeyCredentialRpEntity) -> String {
+    rp.name.clone().unwrap_or(rp.id.clone())
+}
+
+fn format_credential(cred: &Ctap2CredentialData) -> String {
+    cred.user
+        .display_name
+        .clone()
+        .unwrap_or(cred.user.name.clone().unwrap_or("<No username>".into()))
+        .to_string()
+}
+
+async fn enumerate_rps<T: CredentialManagement>(
+    channel: &mut T,
+    pin_provider: &mut Box<dyn PinProvider>,
+) -> Result<Vec<(Ctap2PublicKeyCredentialRpEntity, ByteBuf)>, WebAuthnError> {
+    let (rp, rp_id_hash, total_rps) =
+        handle_retries!(channel.enumerate_rps_begin(pin_provider, TIMEOUT));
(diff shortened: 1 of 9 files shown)


libwebauthn/src/management/credential_management.rs
```diff
@@ -0,0 +1,319 @@
...
+use tracing::info;
+
+#[async_trait]
+pub trait CredentialManagement {
+    async fn get_credential_metadata(
+        &mut self,
+        pin_provider: &mut Box<dyn PinProvider>,
+        timeout: Duration,
+    ) -> Result<Ctap2CredentialManagementMetadata, Error>;
+    async fn enumerate_rps_begin(
```
AlfioEmanueleFresta, line 29: Would it make sense for the higher level APIs (`enumerate_rps`, `enumerate_credentials`) to return `Iterator`s?

The only downside I can think of is that we couldn't easily export the `count`.

No need to address this in this PR.

AlfioEmanueleFresta, line 29: Would it make sense for the higher level APIs (`enumerate_rps`, `enumerate_credentials`) to return `Iterator`s?

(No need to address this in this PR.)

libwebauthn/src/management/credential_management.rs
```diff
@@ -0,0 +1,319 @@
...
+
+            // On success, this is an all-empty Ctap2AuthenticatorConfigResponse
+            handle_errors!(
+                self,
+                self.ctap2_credential_management(&req, timeout).await,
+                uv_auth_used
+            )
+        }?;
+        let metadata = Ctap2CredentialManagementMetadata::new(
+            resp.existing_resident_credentials_count.unwrap(),
```
AlfioEmanueleFresta, line 94: I'm not very comfortable with unwraps on authenticator response fields. IMO we should treat the response as untrusted  - WDYT?

libwebauthn/src/management/credential_management.rs
```diff
@@ -0,0 +1,319 @@
...
+    async fn get_credential_metadata(
+        &mut self,
+        pin_provider: &mut Box<dyn PinProvider>,
+        timeout: Duration,
+    ) -> Result<Ctap2CredentialManagementMetadata, Error>;
+    async fn enumerate_rps_begin(
+        &mut self,
+        pin_provider: &mut Box<dyn PinProvider>,
+        timeout: Duration,
+    ) -> Result<(Ctap2PublicKeyCredentialRpEntity, ByteBuf, u64), Error>;
```
AlfioEmanueleFresta: These output types may not be obvious, I think it's worth creating a struct with named fields.

libwebauthn/examples/cred_management.rs
```diff
@@ -0,0 +1,237 @@
...
+        .display_name
+        .clone()
+        .unwrap_or(cred.user.name.clone().unwrap_or("<No username>".into()))
+        .to_string()
+}
+
+async fn enumerate_rps<T: CredentialManagement>(
+    channel: &mut T,
+    pin_provider: &mut Box<dyn PinProvider>,
+) -> Result<Vec<(Ctap2PublicKeyCredentialRpEntity, ByteBuf)>, WebAuthnError> {
```
AlfioEmanueleFresta: Could we expose Vec<u8> instead of ByteBuf? Thankfully this should be cheap with `into_vec` ([source](https://docs.rs/serde_bytes/latest/src/serde_bytes/bytebuf.rs.html#67-69)).

Context - I'd like for higher level APIs, including the WebAuthn trait, to eventually hide much of the CTAP ugliness: optional fields, protocol internals, serde types. 

This isn't already the case and we have the following tech debt:

1. Currently use Ctap2 models directly within high-level traits, eg.
    ```rust
   pub struct MakeCredentialRequest {
           pub hash: Vec<u8>,
           pub origin: String,
           pub relying_party: Ctap2PublicKeyCredentialRpEntity,
     ```
2. Typealias other types as a shortcut:
    ```

AlfioEmanueleFresta (changes requested):
Thank you so much for your work on this @msirringhaus! 

A few notes, and as this is net new functionality without a stable API, I'll leave the decision as to how much to address here or a separate PR to your judgement.

libwebauthn/src/management/credential_management.rs
msirringhaus: Not sure how nice that would be to use, since each iteration could fail and may needs to be retried. 
Maybe simpler would be to move all the logic of the example inside libwebauthn, and only have a single `enumerate_credentials()`-function that already collects all RPs and all credentials and returns a map with all the content? 
Depends a bit on how low-level we want the API to be. We have the same "problem" with e.g. BioEnrollments.

libwebauthn/src/management/credential_management.rs
msirringhaus: I would agree. We have this however in lots of other places. I'll create a follow up issue to fix this all in one PR, ok?

libwebauthn/src/management/credential_management.rs
msirringhaus: :+1:

libwebauthn/examples/cred_management.rs
msirringhaus: Makes sense

AlfioEmanueleFresta (approved):
Thanks for addressing the feedback.

PR #70, opened 2025-02-12
Fix #68: Implement legacy preview

### LINKED ISSUE #68: Implement Legacy preview of BioEnrollments and CredentialManagement
With the spec version FIDO_2_1_PRE, a "preview" of the commands BioEnrollments and CredentialManagement was introduced.
Sadly, these differ from the final implementation of the commands, but should still be supported.

msirringhaus:


AlfioEmanueleFresta (approved):
Awesome work, thank you!

Do you know of any hardware this could be tested with?

msirringhaus:
Ah, sorry! I completely forgot to mention it: 
My Yubikey 5 has legacy CredentialManagement, and I tested it with that.
However, I don't have a device that uses BioEnrollmentPreview.

PR #72, opened 2025-02-13
Fix #69: Do not unwrap when extracting data from responses

### LINKED ISSUE #69: Do not unwrap when extracting data from responses
> I'm not very comfortable with unwraps on authenticator response fields. IMO we should treat the response as untrusted  - WDYT?

_Originally posted by @AlfioEmanueleFresta in https://github.com/linux-credentials/libwebauthn/pull/67#discussion_r1951669713_

msirringhaus:
There still are a bunch of other unwraps in the code (some of which you can even see in this patch), mostly parsing either CBOR or other binary stuff in the crypto-section.

Not sure, if we want to address them here as well, and if so, how. 
For CBOR-related things, we could basically also just return the serde errors?

### DIFF
--- a/libwebauthn/src/lib.rs
+++ b/libwebauthn/src/lib.rs
@@ -14,9 +14,17 @@ extern crate num_derive;
 extern crate bitflags;
 
 macro_rules! unwrap_field {
-    ($field:expr) => {
-        $field.ok_or(Error::Platform(PlatformError::InvalidDeviceResponse))?
-    };
+    ($field:expr) => {{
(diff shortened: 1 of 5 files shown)


AlfioEmanueleFresta (approved):

msirringhaus:
Added a similar macro for CBOR parsing now. 
There are still a bunch of other unwraps in the code, but I think those should be addressed by another PR. They seem less likely to be triggered any time soon.

AlfioEmanueleFresta (approved):

PR #83, opened 2025-02-21
Implement HMAC secret extension

msirringhaus:
This is working, but I'll mark it as WIP for now, as we may want to incorporate extensions-related issues in here as well.

Noteworthy things here:
- `Ctap2PinUvAuthProtocol` got a new function to create a boxed `PinUvAuthProtocol`-object out of a given protocol version, as handing around the boxed dyn-trait object is quite annoying.
- I extended what will be saved in the channel wrt to auth_token-stuff, as we need the shared_secret, the key_agreement and the used protocol version in order to be able to process the HMAC input AND output. At least, with this, I could de-duplicate `get_uv_auth_token()`. However, there is also potentially a problem here, as we currently always create a shared_secret, which we shouldn't do in the CTAP 2.0 case, see also https://github.com/mozilla/authenticator-rs/issues/341
- Extension input exists in two versions: `HMACGetSecretInput` as given by the user via `GetAssertionResponseExtensions`. This is remapped to `CalculatedHMACGetSecretInput` and `Ctap2GetAssertionRequestExtensions`, respectively. `CalculatedHMACGetSecretInput` holds both the user-input and the info we will send to the device (as we need to cache the user-input somewhere, to calculate the actual hmac-secret input when we have established a shared_secret).
- `HMACGetSecretOutput` works similarly in reverse, but I currently don't rewrap to a new struct. The resulting output will contain the encrypted message from the device and the decrypted info. The encrypted message is marked private, though.

Maybe we can work out a more generic way here, to close some of the open `[Extensions]`-issues.

### DIFF
--- a/libwebauthn/examples/webauthn_extensions_hid.rs
+++ b/libwebauthn/examples/webauthn_extensions_hid.rs
@@ -7,7 +7,7 @@ use rand::{thread_rng, Rng};
 use tracing_subscriber::{self, EnvFilter};
 
 use libwebauthn::ops::webauthn::{
-    GetAssertionRequest, GetAssertionRequestExtensions, MakeCredentialRequest,
+    GetAssertionRequest, GetAssertionRequestExtensions, HMACGetSecretInput, MakeCredentialRequest,
     MakeCredentialsRequestExtensions, UserVerificationRequirement,
 };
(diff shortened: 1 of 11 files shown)


AlfioEmanueleFresta (approved):
Awesome work @msirringhaus, thank you!

PR #85, opened 2025-02-27
PIN fallback, UV retry counter

msirringhaus:


libwebauthn/src/proto/ctap2/model/client_pin.rs
```diff
@@ -135,6 +135,22 @@ impl Ctap2ClientPinRequest {
         }
     }
 
+    pub fn new_get_uv_retries() -> Self {
+        Self {
+            // GetUvRetires never needed the protocol sent. GetPINRetires did for CTAP 2.0
```
iinuwa: nit, small typos:
```suggestion
            // GetUvRetries never needed the protocol sent. GetPINRetries did for CTAP 2.0
```

iinuwa (approved):
only did a quick scan through, but it seems reasonable.

libwebauthn/src/pin.rs
```diff
@@ -46,9 +46,22 @@ impl Default for PinUvAuthToken {
     }
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum PinRequestReason {
+    Direct,
```
AlfioEmanueleFresta: Do you think we could split `Direct` further? From my understanding, we could have:

* `RelyingPartyRequest` - If the request's `userVerification` parameter is set to `required`, or `preferred` and the authenticator  chose to honour the preference, or `credProtect` is set to `userVerificationRequired`;
* `AuthenticatorPolicy` - If `GetInfo` response contains `alwaysUv` (global policy), or any other case (authenticator discretion).

We could them customise the message displayed to the user, eg.:
* `RelyingPartyRequest`: *google.com requires you to use your PIN or biometrics to confirm your identity*.
*  `AuthenticatorPolicy`: *Your device requires you to use your PIN or biometrics to confirm your identity*.

It may be valuable to add some doc comments to each value.

WDYT?

Cc @iinuwa

libwebauthn/src/pin.rs
```diff
@@ -46,9 +46,22 @@ impl Default for PinUvAuthToken {
...
 #[async_trait]
 pub trait PinProvider: Send + Sync {
-    async fn provide_pin(&self, attempts_left: Option<u32>) -> Option<String>;
+    async fn provide_pin(
+        &self,
+        attempts_left: Option<u32>,
+        reason: PinRequestReason,
+    ) -> Option<String>;
+    /// This is for displaying attempts_left only. No direct feedback from the app expected/required.
+    async fn notify_uv_failed(&self, attempts_left: Option<u32>);
```
AlfioEmanueleFresta: Naming nit: Thinking about this further, if we only emit this upon making a new attempt, this should trigger a UI message to prompt the user to try UV again. 

Maybe `prompt_uv_retry`? It would signify more than merely reporting a failure.

Cc @iinuwa

AlfioEmanueleFresta (commented):
The retry works

libwebauthn/src/pin.rs
msirringhaus: Makes sense to me. I'll try to implement that next week!

libwebauthn/src/pin.rs
iinuwa: yup, I agree

libwebauthn/src/pin.rs
iinuwa: That can be a helpful distinction. One caveat: I think that
> RelyingPartyRequest - If the request's `userVerification` parameter is set to ... preferred and the authenticator chose to honour the preference, 

feels more like an `AuthenticatorPolicy`. In this case, it would be inaccurate to say that `google.com` requires UV, since the authenticator decided whether to use UV.

msirringhaus: I think it can be argued both ways, and I honestly have no opinion on which is 'more correct'. In the current version, I added `preferred` also to `RelyingPartyRequest`, simply because we use that in the algorithm before. But we could easily swap it out.

AlfioEmanueleFresta: I see your point @iinuwa. There is some ambiguity, as we won't know for certain if the authenticator honoured the `preferred` request, or chose to require UV for a different internal reason.

IMO the best we can do is to document that our reason enum will give priority to the authenticator request. The message to the user may need to be worded accordingly. Ie. *google.com **requested** that you use your PIN* (always factually true), rather than *google.com **requires** you to use your PIN* (inaccurate).

I think this would be more accurate than telling users they're being asked for UV as the authenticator required it - as it may well be the authenticator would never have done so had the RP not asked for it.

AlfioEmanueleFresta:
Thank you very much for this work @msirringhaus!

Merging this as it's got two approvals; we can follow up on the reason enum discussion separately if needed.

PR #89, opened 2025-03-11
Implement state update and add PresenceRequired status

msirringhaus:
Couple of things going on here:
- Channel-creation now also returns a state update receiver that the user has to poll (currently, the spawn is outside of libwebauthn, but we could move it inside as well)
- The state update channel completely replaces the PinProvider/UvProvider trait. It will transport both generic updates ("Touch your device"), as well as provide a way (via oneshot-channels) for the application to send back the PIN.
- Introduced `PresenceRequired`-state, to tell the user when to touch the device. Some weird devices might behave differently, though, and we can't know this.
- In order to only send one `PresenceRequired`-state update, we have to also cache, if builtin UV or PIN was used in `AuthTokenData`. For UV we have to send out the update BEFORE we establish the auth token, for PIN, AFTER. And only in some cases (FIDO2 MakeCredential and GetAssertion, but not for CredentialManagement, etc.).
- Removed a few functions from `Device` that weren't really used anywhere, and moved `wink()` to the channel (mostly to avoid the additional creation of a very short lived Channel, that was thrown away after `Device::wink()` returned). 
- `Channel::wink()` didn't wait for the reply from the device.

I have tested most examples with this, except for webauthn_cable.rs

### DIFF
--- a/libwebauthn/src/lib.rs
+++ b/libwebauthn/src/lib.rs
@@ -6,6 +6,7 @@ pub mod proto;
 pub mod transport;
 pub mod u2f;
 pub mod webauthn;
+use tokio::sync::oneshot;
 
 #[macro_use]
 extern crate num_derive;
@@ -26,12 +27,27 @@ macro_rules! unwrap_field {
         }
     }};
 }
+use pin::PinRequestReason;
 pub(crate) use unwrap_field;
 
 #[derive(Debug)]
 pub enum Transport {
     Usb,
     Ble,
+}
+
+#[derive(Debug)]
+pub enum StateUpdate {
+    UvRetry {
+        attempts_left: Option<u32>,
+    },
+    /// Oneshot channel
+    PinRequired {
+        reply_to: oneshot::Sender<String>,
+        reason: PinRequestReason,
+        attempts_left: Option<u32>,
+    },
+    PresenceRequired,
 }
 
 pub fn available_transports() -> Vec<Transport> {
--- a/libwebauthn/examples/authenticator_config_hid.rs
+++ b/libwebauthn/examples/authenticator_config_hid.rs
@@ -3,13 +3,15 @@ use std::fmt::Display;
 use std::time::Duration;
 
 use libwebauthn::management::AuthenticatorConfig;
-use libwebauthn::pin::{UvProvider, StdinPromptPinProvider};
+use libwebauthn::pin::PinRequestReason;
 use libwebauthn::proto::ctap2::{Ctap2, Ctap2GetInfoResponse};
 use libwebauthn::transport::hid::list_devices;
 use libwebauthn::transport::Device;
 use libwebauthn::webauthn::Error as WebAuthnError;
+use libwebauthn::StateUpdate;
 use std::io::{self, Write};
 use text_io::read;
+use tokio::sync::mpsc::Receiver;
 use tracing_subscriber::{self, EnvFilter};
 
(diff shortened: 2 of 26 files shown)


libwebauthn/src/lib.rs
```diff
@@ -34,6 +36,20 @@ pub enum Transport {
     Ble,
 }
 
+#[derive(Debug)]
+pub enum StateUpdate {
```
AlfioEmanueleFresta: State sounds overly generic. These are UX hints for UV and UP, maybe `UxUpdate`?

libwebauthn/src/lib.rs
```diff
@@ -34,6 +36,20 @@ pub enum Transport {
...
+    UvRetry {
+        attempts_left: Option<u32>,
+    },
+    /// Oneshot channel
+    PinRequired {
```
AlfioEmanueleFresta: Could we add some simple docs to each of these variants? Here it would be helpful to explain the caller must use `reply_to`, and that a timeout may occur.

libwebauthn/src/lib.rs
```diff
@@ -34,6 +36,20 @@ pub enum Transport {
...
+        reply_to: oneshot::Sender<String>,
```
AlfioEmanueleFresta: I'm wondering if it makes sense to make each variant its own struct, eg.:


```rust
pub enum UxUpdate {
  UvRetry(UvRetryUpdate),
  PinRequired(PinRequiredUpdate),
  PresenceRequired(PresenceRequiredUpdate)
}
```

If we did, we could have the channel be an internal property, including any types used within the channel, and offer simple methods agnostic of internal structures. Eg.

```rust
pub struct PinRequiredUpdate {
  reply: oneshot::Sender<Option<String>>,
  pub reason: PinRequestReason,
  pub attempts_left: Option<u32>,
```

And then, a possibly simpler:

```rust
impl PinRequiredUpdate {
  /// The user is making a new attempt at entering a PIN.
  pub fn provide_pin(&self, pin: &str) { 
    self.reply_to.send(Some(pin.to_owned())
  }
  
  /// The user cancels the PIN entry, without making an attempt.
  pub async fn cancel() { 
    self.reply_to.send(None)
  }
}
```

WDYT?

AlfioEmanueleFresta (approved):
Thanks for your work on this @msirringhaus, I love the new API!

A few follow up comments, feel free to address them here or as a subsequent PR as per your preference.

libwebauthn/src/lib.rs
msirringhaus: I was going for overly generic :D But `UxUpdate` does fit really well, indeed!

libwebauthn/src/lib.rs
msirringhaus: For `PinRequired`, this definitely makes sense, yes! Not sure about the other two variants, though. `PresenceRequired` has (currently) no data at all, so I wouldn't include an empty struct there. `UvRetry` only has one `Option`, so it's a debate between efficiency (one additional redirect) and consistency. I'm kind of 50:50 on that one.

msirringhaus:
Pushed another commit to address feedback.

Two things to note:
- I had to not wait for an answer from the virtual device on `wink()`, because apparently the solokey doesn't send one and then the tests hang there
- I have not wrapped `UvRetry` in a struct, but I can still do so, if wished.

AlfioEmanueleFresta (approved):
Thank you, LGTM!

PR #90, opened 2025-03-21
Implement Prf

msirringhaus:
This is split into two commits:
The first one implements the translation layer from Ctap2 to Webauthn for MakeCredentials (which was, until now a No-op, and thus not implemented)

The second one then implements PRF on top of that (and the translation layer for GetAssertion).

The only 'ugly' bit to me is the U2F function (see the inline comment).

NOTE: This should be followed up by another PR implementing PreFlight for CTAP2. I think we only have it for U2F at the moment, but we need a filtered allow_list.

### DIFF
--- a/libwebauthn/examples/prf_test.rs
+++ b/libwebauthn/examples/prf_test.rs
@@ -1,0 +1,194 @@
+use std::collections::HashMap;
+use std::convert::TryInto;
+use std::error::Error;
+use std::io::{self, Write};
+use std::time::Duration;
+
+use libwebauthn::transport::hid::channel::HidChannel;
+use libwebauthn::UxUpdate;
+use rand::{thread_rng, Rng};
+use serde_bytes::ByteBuf;
+use text_io::read;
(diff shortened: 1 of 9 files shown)


AlfioEmanueleFresta (approved):
Awesome, thank you! 

Were you able to verify the PRF outputs match those produced by a known-good implementation? I'm happy to run some tests.

msirringhaus:
Good point. I did not.

msirringhaus:
I confirmed the (simplest) output from Firefox on https://demo.yubico.com/webauthn-developers with this PR and the output matches.

I'm wondering, if I should cook up some additional test binary, that let's one insert the credential_id and the first PRF input and calculate the output.
Then one could register with whatever browser one wants on the above demo-page, copy over those two inputs and run it against our implementation, to verify. wdyt?

msirringhaus:
Ok, added a small test example. Verified it now with both Firefox and Chromium.

AlfioEmanueleFresta (approved):

PR #93, opened 2025-04-23
More extensions

msirringhaus:
1. Credential protection policy: 
    a) Stop exposing `ctap_types` in our API and use our own implementation for the webauthn-layer (the CTAP layer still uses `ctap_types`)
    b) Also support `enforce_policy`, which we can only do from the inside, as we need `Ctap2GetInfoResponse` for that. Error out, if it can't be enforced.

2. Support `credProps` extension. 
    a) Pretty straight forward, except that CTAP 2.0 devices are allowed to create discoverable credentials even if they are not requested. CTAP 2.1 devices are not allowed to do that anymore. So we need `Ctap2GetInfoResponse` once again to decide

3. Switch from LargeBlobKeys extension to LargeBlob extension.
    a) LargeBlob extension has "Preferred"-mode as well, so we need `Ctap2GetInfoResponse` again for deciding, if we can request it or not.
    b) LargeBlobKey-requests can easily be mapped into LargeBlob requests, by simply using LargeBlob `support = "required"`, so I'm not supporting both in the API.
    c) `GetAssertionLargeBlobExtension::Write()` is not yet supported, as we need the corresponding CTAP commands to store large blobs for that, which we don't have yet.

### DIFF
--- a/libwebauthn/examples/prf_test.rs
+++ b/libwebauthn/examples/prf_test.rs
@@ -154,8 +154,8 @@ async fn run_success_test(
         allow: vec![credential.clone()],
         user_verification: UserVerificationRequirement::Preferred,
         extensions: Some(GetAssertionRequestExtensions {
-            cred_blob: None,
             hmac_or_prf,
+            ..Default::default()
         }),
         timeout: TIMEOUT,
     };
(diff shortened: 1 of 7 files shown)


AlfioEmanueleFresta (approved):
Thank you!

PR #96, opened 2025-05-04
Fix #95, correct key ordering for CBOR encoding

### LINKED ISSUE #95: Non-canonical CBOR encodings cause MissingParameter errors with Titan Security Key
I'm trying to use a USB-C/NFC [Titan Security Key](https://en.wikipedia.org/wiki/Titan_Security_Key) v2 (K52T), but for example webauthn_hid  panics with a Ctap(MissingParameter) error.

When decoding the CBOR data in trace/debug logs from python-fido2 and libwebauthn I noticed the keys of the PublicKeyCredentialParameters are sorted differently.

In python-fido2:

```
    {
      "alg": -7,
      "type": "public-key"
    }

```
In libwebauthn:

```
    {
      "type": "public-key",
      "alg": -7
    }

```

It seems the CBOR encoding generated by libwebauthn doesn't follow the CTAP2 canonical CBOR encoding form rules in the CTAP spec.  

> 
> The keys in every map MUST be sorted lowest value to highest. The sorting rules are:
> 
>* If the major types are different, the one with the lower value in numerical order sorts earlier.
> 
>* If two keys have different lengths, the shorter one sorts earlier;
> 
>* If two keys have the same length, the one with the lower value in (byte-wise) lexical order sorts earlier.
> 

The string "alg" is shorter than "type" which means "alg" should be sorted before "type", which isn't the case in libwebauthn.

(I also verified I can trigger the same error (MISSING_PARAMETER) by changing the sort order in python-fido2.)

Edit:  Apparently serde_cbor respects the struct field order, and at least the following patch is needed for webauthn_hid to succeed on Titan Security Key.

```
diff --git a/libwebauthn/src/proto/ctap2/model.rs b/libwebauthn/src/proto/ctap2/model.rs
index 63dabab..12b68d1 100644
--- a/libwebauthn/src/proto/ctap2/model.rs
+++ b/libwebauthn/src/proto/ctap2/model.rs
@@ -139,8 +139,8 @@ impl From<&Ctap1Transport> for Ctap2Transport {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Ctap2PublicKeyCredentialDescriptor {
-    pub r#type: Ctap2PublicKeyCredentialType,
     pub id: ByteBuf,
+    pub r#type: Ctap2PublicKeyCredentialType,
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub transports: Option<Vec<Ctap2Transport>>,
@@ -156,11 +156,11 @@ pub enum Ctap2COSEAlgorithmIdentifier {
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct Ctap2CredentialType {
-    #[serde(rename = "type")]
-    pub public_key_type: Ctap2PublicKeyCredentialType,
-
     #[serde(rename = "alg")]
     pub algorithm: Ctap2COSEAlgorithmIdentifier,
+
+    #[serde(rename = "type")]
+    pub public_key_type: Ctap2PublicKeyCredentialType,
 }
 
 impl Default for Ctap2CredentialType {

```

AlfioEmanueleFresta:
See #95.

No trivial root-cause fix due to serde_cbor, but thankfully the CBOR data structures with string keys in libwebauthn are limited.

Added tests to prevent regressions.

## Test vectors

As decoded by [cbor.me](https://cbor.me).

### Ctap2CredentialType

```
A2                         # map(2)
   63                      # text(3)
      616C67               # "alg"
   26                      # negative(6)
   64                      # text(4)
      74797065             # "type"
   6A                      # text(10)
      7075626C69632D6B6579 # "public-key"
```

### Ctap2PublicKeyCredentialDescriptor

```
A2                         # map(2)
   62                      # text(2)
      6964                 # "id"
   41                      # bytes(1)
      42                   # "B"
   64                      # text(4)
      74797065             # "type"
   6A                      # text(10)
      7075626C69632D6B6579 # "public-key"
```

### DIFF
--- a/libwebauthn/src/proto/ctap2/model.rs
+++ b/libwebauthn/src/proto/ctap2/model.rs
@@ -139,8 +139,8 @@ impl From<&Ctap1Transport> for Ctap2Transport {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Ctap2PublicKeyCredentialDescriptor {
-    pub r#type: Ctap2PublicKeyCredentialType,
     pub id: ByteBuf,
+    pub r#type: Ctap2PublicKeyCredentialType,
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub transports: Option<Vec<Ctap2Transport>>,
@@ -156,11 +156,11 @@ pub enum Ctap2COSEAlgorithmIdentifier {
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct Ctap2CredentialType {
-    #[serde(rename = "type")]
-    pub public_key_type: Ctap2PublicKeyCredentialType,
-
     #[serde(rename = "alg")]
     pub algorithm: Ctap2COSEAlgorithmIdentifier,
+
+    #[serde(rename = "type")]
+    pub public_key_type: Ctap2PublicKeyCredentialType,
 }
 
 impl Default for Ctap2CredentialType {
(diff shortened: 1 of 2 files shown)


kalvdans (approved):

libwebauthn/src/proto/ctap2/model.rs
```diff
@@ -139,8 +139,8 @@ impl From<&Ctap1Transport> for Ctap2Transport {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Ctap2PublicKeyCredentialDescriptor {
-    pub r#type: Ctap2PublicKeyCredentialType,
     pub id: ByteBuf,
+    pub r#type: Ctap2PublicKeyCredentialType,
```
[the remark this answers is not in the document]
AlfioEmanueleFresta: Sadly this is nonsense, the change does exactly what the comment states, and the code suggestion reverses it.

libwebauthn/src/proto/ctap2/model.rs
```diff
@@ -205,3 +205,40 @@ pub enum Ctap2UserVerificationOperation {
...
+
+    #[test]
+    /// Verify CBOR serialization conforms to CTAP canonical standard, including ordering (see #95) 
+    pub fn credential_type_field_serialization() {
+        let credential_type = Ctap2CredentialType {
+            algorithm: Ctap2COSEAlgorithmIdentifier::ES256,
+            public_key_type: Ctap2PublicKeyCredentialType::PublicKey,
+        };
+        let serialized = serde_cbor::to_vec(&credential_type).unwrap();
+        // Known good, verified by hand with cbor.me playground
```
msirringhaus, line 226: Maybe add the cbor.me-output here to the comment:
```
A2                         # map(2)
   63                      # text(3)
      616C67               # "alg"
   26                      # negative(6)
   64                      # text(4)
      74797065             # "type"
   6A                      # text(10)
      7075626C69632D6B6579 # "public-key"
```

I often find that helpful.
Same below.

msirringhaus (approved):
LGTM!

libwebauthn/src/proto/ctap2/model.rs
AlfioEmanueleFresta: Sorry, I only saw this after merging. The output made it to the commit message, so hopefully this is accessible via git blame.

msirringhaus: yeah, that's fine.

PR #98, opened 2025-05-05
Hybrid transport: Switch to event-driven discovery

AlfioEmanueleFresta:
## Changes

* Switches from a periodic polling approach, to an event-driven discovery. It uses btleplug's event system to wait for a BLE advertisement. This is guaranteed to occur after the QR code is generated, making this safe.
* Fixes iOS discovery limitation, by removing UUID filter for BlueZ device discovery. This seems to be a limitation in BlueZ (it [wouldn't be the first](https://github.com/deviceplug/btleplug/issues/289)), that I could not find a workaround for. It should have no significant downside, however, as discovered devices that don't include the relevant service data are ignored (we do not attempt connecting to them). 

## Testing

Tested on an iPhone, and Android 15. Example `webauthn_cable` succeeds consistently.

### DIFF
--- a/libwebauthn/src/transport/ble/btleplug/manager.rs
+++ b/libwebauthn/src/transport/ble/btleplug/manager.rs
@@ -78,10 +78,10 @@ async fn on_peripheral_service_data(
 }
 
 #[instrument(level = Level::DEBUG, skip_all)]
-/// Starts a discovery for devices with specific service data
+/// Starts a discovery for devices advertising service data on any of the provided UUIDs
 pub async fn start_discovery_for_service_data(
-    uuids: Vec<Uuid>,
-) -> Result<impl Stream<Item = (Peripheral, Vec<u8>)>, Error> {
+    uuids: &[Uuid],
+) -> Result<impl Stream<Item = (Peripheral, Vec<u8>)> + use<'_>, Error> {
     let adapter = get_adapter().await?;
     let scan_filter = ScanFilter::default();
 
@@ -93,11 +93,9 @@ pub async fn start_discovery_for_service_data(
         .or(Err(Error::ConnectionFailed))?;
 
     let stream = events.filter_map({
-        let adapter = adapter.clone();
-        let uuids = uuids.clone();
         move |event| {
             let adapter = adapter.clone();
-            let uuids = uuids.clone();
+            let uuids = uuids.to_vec();
             async move {
                 // trace!(?event);
                 match event {
--- a/libwebauthn/src/transport/ble/btleplug/manager.rs
+++ b/libwebauthn/src/transport/ble/btleplug/manager.rs
@@ -2,10 +2,11 @@ use std::collections::HashMap;
 
 use btleplug::api::bleuuid::uuid_from_u16;
 use btleplug::api::{
-    Central as _, Manager as _, Peripheral as _, PeripheralProperties, ScanFilter,
+    Central as _, CentralEvent, Manager as _, Peripheral as _, PeripheralProperties, ScanFilter,
 };
-use btleplug::platform::{Adapter, Manager, Peripheral};
-use tracing::{debug, info, instrument, Level};
+use btleplug::platform::{Adapter, Manager, Peripheral, PeripheralId};
+use futures::{Stream, StreamExt};
+use tracing::{debug, info, instrument, trace, warn, Level};
 use uuid::Uuid;
 
 use super::device::FidoEndpoints;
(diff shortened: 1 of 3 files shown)


libwebauthn/src/transport/ble/btleplug/manager.rs
```diff
@@ -50,17 +51,66 @@ impl SupportedRevisions {
...
+        "Ignoring periperal as it doesn't have service data for desired UUID"
+    );
+    None
+}
+
 #[instrument(level = Level::DEBUG, skip_all)]
-pub async fn start_discovery(uuids: &[Uuid]) -> Result<(), Error> {
+/// Starts a discovery for devices with specific service data
+pub async fn start_discovery_for_service_data(
+    uuids: Vec<Uuid>,
```
msirringhaus: Why is this a `Vec<>` now, instead of a reference?

libwebauthn/src/transport/ble/btleplug/manager.rs
```diff
@@ -50,17 +51,66 @@ impl SupportedRevisions {
...
         .await
-        .or(Err(Error::ConnectionFailed))
+        .or(Err(Error::ConnectionFailed))?;
+
+    let stream = events.filter_map({
+        let adapter = adapter.clone();
+        let uuids = uuids.clone();
+        move |event| {
+            let adapter = adapter.clone();
+            let uuids = uuids.clone();
```
msirringhaus: Do you actually need to clone uuids and adapter, if you only access them by reference below?

libwebauthn/src/transport/cable/qr_code_device.rs
```diff
@@ -185,50 +184,48 @@ impl CableQrCodeDevice<'_> {
     }
 
     async fn await_advertisement(&self) -> Result<(FidoDevice, DecryptedAdvert), Error> {
-        btleplug::manager::start_discovery(&[
-            Uuid::parse_str(CABLE_UUID_FIDO).unwrap(),
+        let stream = btleplug::manager::start_discovery_for_service_data(vec![
```
msirringhaus: As mentioned above, I feel like this could still be a `&[]` instead of creating a vec.

libwebauthn/src/transport/cable/qr_code_device.rs
```diff
@@ -185,50 +184,48 @@ impl CableQrCodeDevice<'_> {
...
+            debug!(
+                ?device,
+                ?decrypted,
+                "Successfully decrypted advertisement from device"
+            );
+
+            return Ok((device, advert));
         }
+
+        unreachable!()
```
msirringhaus: Is this really unreachable? Can't `stream.as_mut().next()` ever return `None`? (Due to timeouts, HW removed, etc.)

libwebauthn/src/transport/cable/qr_code_device.rs
```diff
@@ -185,50 +184,48 @@ impl CableQrCodeDevice<'_> {
     }
 
     async fn await_advertisement(&self) -> Result<(FidoDevice, DecryptedAdvert), Error> {
-        btleplug::manager::start_discovery(&[
-            Uuid::parse_str(CABLE_UUID_FIDO).unwrap(),
+        let stream = btleplug::manager::start_discovery_for_service_data(vec![
             Uuid::parse_str(CABLE_UUID_GOOGLE).unwrap(),
+            Uuid::parse_str(CABLE_UUID_FIDO).unwrap(),
```
msirringhaus, line 188: Is the order of the UUIDs here important? If so, maybe add a comment to avoid accidental reordering.

msirringhaus (commented):
I've left some minor comments

libwebauthn/src/transport/cable/qr_code_device.rs
AlfioEmanueleFresta: Indeed not, good catch! Fixed.

libwebauthn/src/transport/cable/qr_code_device.rs
AlfioEmanueleFresta: You're right, sorry I didn't clean this up before the PR. Fixed.

libwebauthn/src/transport/ble/btleplug/manager.rs
AlfioEmanueleFresta: I should have cleaned this up, I removed the duplicate cloning. I do need one though because of the surviving `async move` scope, and neither types being `Copy`.

libwebauthn/src/transport/ble/btleplug/manager.rs
AlfioEmanueleFresta: I should have reverted this, sorry! Fixed.

libwebauthn/src/transport/cable/qr_code_device.rs
AlfioEmanueleFresta: Good point. The order is not important here, the Google UUID is only for caBLE v2 backwards compatibility. I've added some comments explaining this.

AlfioEmanueleFresta:
Thanks @msirringhaus! Addressed your comments.

msirringhaus (approved):
Thank you!

PR #99, opened 2025-05-08
Split extensions into unsigned extensions and signed extensions inside AuthenticatorData

msirringhaus:
- `impl Serialize for AttestedCredentialData` and `impl<T> Serialize for AuthenticatorData<T>` actually doesn't work the way it was intended, because it needs to be wrapped inside some other data type, that we don't have. So I moved that code to a stand-alone function that returns the `Vec<u8>` directly.
- Some things are not yet implemented, like appid, largeblob-write.


AlfioEmanueleFresta (approved):

msirringhaus:
I only renamed one function, which was quite generic and unclear.

Issue #66, opened 2025-02-09
Ignore unknown CTAP response fields


AlfioEmanueleFresta:
New versions of the specifications frequently add new fields. Ignoring new fields allow previous versions of the library not to fail when used with newer devices. See #65.

We may need to fork serde_indexed to add this functionality.

PR #100, opened 2025-05-11
Hybrid transport: Add state-assisted transactions

AlfioEmanueleFresta:
## Changes

* Updates internals of CableKnownDeviceInfoStore, to make it easier to use and thread-safe.
* Adds parsing of caBLE update messages. This includes a Google-specific key (999). Due to a limitation in serde-indexed, parsing of this structure has to be done field-by-field with `serde_cbor`. To be fixed properly with #66.
* Connect to known device from saved state.
* Verifies handshake signature. Note this currently only works for linking information received on QR-initiated transactions. The protocol doesn't seen to support subsequent updates, as the ephemeral QR-code identity key is needed for verification.

### DIFF
--- a/libwebauthn/src/transport/cable/advertisement.rs
+++ b/libwebauthn/src/transport/cable/advertisement.rs
@@ -19,8 +19,8 @@ pub(crate) struct DecryptedAdvert {
     pub encoded_tunnel_server_domain: u16,
 }
 
-impl From<&[u8]> for DecryptedAdvert {
-    fn from(plaintext: &[u8]) -> Self {
+impl From<[u8; 16]> for DecryptedAdvert {
+    fn from(plaintext: [u8; 16]) -> Self {
         let mut nonce = [0u8; 10];
         nonce.copy_from_slice(&plaintext[1..11]);
         let mut routing_id = [0u8; 3];
@@ -70,7 +70,7 @@ pub(crate) async fn await_advertisement(
         };
         trace!(?decrypted);
 
-        let advert = DecryptedAdvert::from(decrypted.as_slice());
+        let advert = DecryptedAdvert::from(decrypted);
         debug!(
             ?device,
             ?decrypted,
--- a/libwebauthn/src/transport/cable/advertisement.rs
+++ b/libwebauthn/src/transport/cable/advertisement.rs
@@ -1,0 +1,90 @@
+use ::btleplug::api::Central;
+use futures::StreamExt;
+use std::pin::pin;
+use tracing::{debug, trace, warn};
+use uuid::Uuid;
+
+use crate::transport::ble::btleplug::{self, FidoDevice};
+use crate::transport::cable::crypto::trial_decrypt_advert;
+use crate::webauthn::{Error, TransportError};
+
+const CABLE_UUID_FIDO: &str = "0000fff9-0000-1000-8000-00805f9b34fb";
+const CABLE_UUID_GOOGLE: &str = "0000fde2-0000-1000-8000-00805f9b34fb";
+
+#[derive(Debug)]
+pub(crate) struct DecryptedAdvert {
+    pub plaintext: [u8; 16],
+    pub nonce: [u8; 10],
+    pub routing_id: [u8; 3],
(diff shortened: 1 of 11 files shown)


libwebauthn/src/transport/cable/tunnel.rs
```diff
@@ -513,7 +555,72 @@ async fn connection_recv_initial(
     Ok(initial_message.info.to_vec())
 }
 
+async fn connection_recv_update(message: &[u8]) -> Result<Option<CableLinkingInfo>, Error> {
```
msirringhaus, line 590: It's more or less equivalent, but I think we could use a custom deserialize-implementation for this, like done here: https://github.com/mozilla/authenticator-rs/blob/ctap2-2021/src/ctap2/commands/get_assertion.rs#L734

libwebauthn/src/transport/cable/tunnel.rs
```diff
@@ -513,7 +555,72 @@ async fn connection_recv_initial(
...
+    let update_message: BTreeMap<Value, Value> = match serde_cbor::from_slice(&message) {
+        Ok(update_message) => update_message,
+        Err(e) => {
+            error!(?e, "Failed to decode update message");
+            return Err(Error::Transport(TransportError::ConnectionFailed));
+        }
+    };
+
+    let Some(Value::Map(linking_info_map)) = update_message.get(&Value::Integer(0x01)) else {
+        warn!("Epty linking info map");
```
msirringhaus: Typo `Epty`

msirringhaus (commented):
Only small comments so far

libwebauthn/src/transport/cable/tunnel.rs
AlfioEmanueleFresta: Agreed, I think we'll have to do this more widely to fix #66 very soon. OK to do this as part of that task?

AlfioEmanueleFresta:
Connecting to a known device is now working on Android.

I'll add handshake verification next as part of this PR.

libwebauthn/src/transport/cable/advertisement.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub plaintext: [u8; 16],
+    pub nonce: [u8; 10],
+    pub routing_id: [u8; 3],
+    pub encoded_tunnel_server_domain: u16,
+}
+
+impl From<&[u8]> for DecryptedAdvert {
+    fn from(plaintext: &[u8]) -> Self {
+        let mut nonce = [0u8; 10];
+        nonce.copy_from_slice(&plaintext[1..11]);
```
msirringhaus, line 25: Do we need bound checks here for plaintext?

libwebauthn/src/transport/cable/tunnel.rs
```diff
@@ -94,6 +96,24 @@ struct CableInitialMessage {
...
+    pub contact_id: Vec<u8>,
+    /// Used by the authenticator to identify the client platform
+    pub link_id: Vec<u8>,
+    /// Shared secret between authenticator and client platform
+    pub link_secret: Vec<u8>,
+    /// Authenticator's public key, X9.62 uncompressed format
+    pub authenticator_public_key: Vec<u8>,
+    /// User-friendly name of the authenticator
+    pub authenticator_name: String,
+    /// HMAC of the handshake hash (Noise's channel binding value) usinng the
```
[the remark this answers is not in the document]
msirringhaus: For once, copilot is right about something

libwebauthn/src/transport/cable/tunnel.rs
msirringhaus: Sure

msirringhaus (approved):
Small nitpicks. I currently can't judge the overall spec compliance, but I think you know what you are doing :)

libwebauthn/src/transport/cable/advertisement.rs
AlfioEmanueleFresta: Good call, this was old code but I've changed it to use a fixed-size array.

PR #104, opened 2025-05-22
Parse unsigned extensions, even if no signed extensions are present

msirringhaus:
Some unsigned extensions have to be present, even if no signed extensions are present. Thus we have to decouple them.

And rename to camelCase seemed to make `Id` out of `ID`, which is needed for JSON-parsing. So we manually rename the enums in `CredentialProtectionPolicy` now, just to be sure.


AlfioEmanueleFresta (approved):
Thank you!

PR #105, opened 2025-05-23
Add NFC transport

mikma:
This draft PR adds an NFC transport with two backends: PC/SC and libnfc. The backends are enabled using features: pcsc and libnfc.

The NFC transport is not complete yet, support for ctap1 is missing, and also support for NFCCTAP_GETRESPONSE, which seems to be optional.

When finished this PR should fix #5.

I have tested the transport with two NFC readers:

- DFRobot NFC module based on PN532 and connected to a USB to serial adapter and supported by libnfc.
- ACR122U-A9 with PC/SC driver (libacsccid1).

And I have used the USB-C Titan Security Key (K52T) with both readers.

What do you think?

PR #107, opened 2025-05-23
Use thiserror for errors

AlfioEmanueleFresta:
* Reduces some boilerplate
* Used in  #105 - allow making dependency non-optional, rather than for `nfc` feature only
* I plan to use this in new CBOR serialization and deserialization, as I would like to encapsulate underlying errors (more easily)

### DIFF
--- a/libwebauthn/src/transport/error.rs
+++ b/libwebauthn/src/transport/error.rs
@@ -1,6 +1,6 @@
 pub use crate::proto::CtapError;
 
-#[derive(thiserror::Error, Debug, PartialEq)]
+#[derive(thiserror::Error, Debug, Copy, Clone, PartialEq)]
 pub enum PlatformError {
     #[error("pin too short")]
     PinTooShort,
@@ -18,7 +18,7 @@ pub enum PlatformError {
     SyntaxError,
 }
 
-#[derive(thiserror::Error, Debug, PartialEq)]
+#[derive(thiserror::Error, Debug, Copy, Clone, PartialEq)]
 pub enum TransportError {
     #[error("connection failed")]
     ConnectionFailed,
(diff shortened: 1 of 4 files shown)


libwebauthn/src/transport/error.rs
```diff
@@ -1,81 +1,59 @@
 pub use crate::proto::CtapError;
 
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(thiserror::Error, Debug, PartialEq)]
```
msirringhaus: I think we should keep at least `Clone` in there.

libwebauthn/src/transport/error.rs
```diff
@@ -1,81 +1,59 @@
...
-impl std::error::Error for PlatformError {}
-
-impl std::fmt::Display for PlatformError {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{:?}", self)
-    }
-}
-
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(thiserror::Error, Debug, PartialEq)]
```
msirringhaus: Same here. `Clone` is very useful for the API.

libwebauthn/src/transport/error.rs
```diff
@@ -1,81 +1,59 @@
...
-impl std::error::Error for TransportError {}
-
-impl std::fmt::Display for TransportError {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{:?}", self)
-    }
-}
-
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(thiserror::Error, Debug, PartialEq)]
```
msirringhaus: Dito

AlfioEmanueleFresta:
Thanks @msirringhaus, good points. Addressed, good to merge?

msirringhaus (approved):
Yes, thank you!

PR #108, opened 2025-05-25
Add Hybrid variant to Ctap2Transport enum 

AlfioEmanueleFresta:
* Fixes casing of CTAP1 and CTAP2 transport enums, as per Rust's official naming conventions. Note this applies even to acronyms - examples are IpAddr, TcpStream, Utf8Error from std.
* Adds Hybrid transport variant.


libwebauthn/src/proto/ctap2/model/credential_protection.rs
```diff
@@ -0,0 +1,22 @@
+use minicbor::{Decode, Encode};
```
msirringhaus: Is this file supposed to be part of this PR? 
The `CredentialProtectionPolicy` isn't used anywhere, it seems.

AlfioEmanueleFresta: :facepalm:

msirringhaus (approved):

PR #109, opened 2025-05-25
Make the HID channel safer (fix #103)

### LINKED ISSUE #103: HidChannel panics when USB device is unplugged
If you unplug the device while trying to send something to it, the library panics. We'll need to clean turn this map this error instead:

https://github.com/linux-credentials/libwebauthn/blob/master/libwebauthn/src/transport/hid/channel.rs#L267

I found this while testing linux-credentials/linux-webauthn-platform-api#16.

AlfioEmanueleFresta:
Fixes #103.

Also protect against mutex poisoning.

### DIFF
--- a/libwebauthn/src/transport/hid/channel.rs
+++ b/libwebauthn/src/transport/hid/channel.rs
@@ -246,7 +246,10 @@ impl<'d> HidChannel<'d> {
     pub async fn hid_send(&self, msg: &HidMessage) -> Result<(), Error> {
         match &self.open_device {
             OpenHidDevice::HidApiDevice(hidapi_device) => {
-                let guard = hidapi_device.lock().unwrap();
+                let Ok(guard) = hidapi_device.lock() else {
+                    warn!("Poisoned lock on HID API device");
+                    return Err(Error::Transport(TransportError::ConnectionLost));
+                };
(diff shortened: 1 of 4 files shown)


libwebauthn/src/transport/hid/channel.rs
```diff
@@ -440,7 +448,10 @@ impl Channel for HidChannel<'_> {
         let cid = self.init.cid;
         debug!({ cid }, "Sending APDU request");
         trace!(?request);
-        let apdu_raw = request.raw_long().unwrap();
+        let Ok(apdu_raw) = request.raw_long() else {
+            warn!("Failed to serialize APDU request");
+            return Err(Error::Transport(TransportError::InvalidFraming));
```
msirringhaus: If we merge the `thiserror`-PR before this one, we could probably include the actual `IOError` here in the `TransportError` instead of throwing the original error away and remapping it to `InvalidFraming`?

AlfioEmanueleFresta: Good point, added. Unfortunately IOError is missing a few traits we require (Clone, PartialEq, etc.) so I've added std::io::ErrorKind which might give us most of what we need.

PR #110, opened 2025-05-25
Allow unknown CBOR fields (ref #66)

### LINKED ISSUE #66: Ignore unknown CTAP response fields

AlfioEmanueleFresta:
* Part fixes #66 for unknown CBOR structures decoded via DeserializeIndexed from serde-indexed.
* Remaining to allow unknown enum variants, to be fixed in a later PR.
 
Depends on https://github.com/trussed-dev/serde-indexed/pull/19.

### DIFF
--- a/libwebauthn/src/proto/ctap2/cbor/serde.rs
+++ b/libwebauthn/src/proto/ctap2/cbor/serde.rs
@@ -1,0 +1,75 @@
+use serde::Serialize;
+
+#[derive(thiserror::Error, Debug)]
+pub enum CborError {
+    #[error("serde_cbor serialization error: {0}")]
+    SerdeCbor(#[from] serde_cbor::Error),
+}
+
+impl PartialEq for CborError {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (CborError::SerdeCbor(e1), CborError::SerdeCbor(e2)) => {
+                e1.to_string() == e2.to_string()
+            }
+        }
+    }
+}
+
+pub(crate) trait CborSerialize {
+    fn to_vec(&self) -> Result<Vec<u8>, CborError>;
+}
+
+impl<T> CborSerialize for T
+where
+    T: Serialize,
+{
+    fn to_vec(&self) -> Result<Vec<u8>, CborError> {
+        serde_cbor::ser::to_vec(self).map_err(CborError::from)
+    }
+}
+
+pub(crate) trait CborDeserialize<T>: Sized + serde::de::DeserializeOwned {
+    fn from_reader<R: std::io::Read>(reader: R) -> Result<Self, CborError>;
+    fn from_slice(slice: &[u8]) -> Result<Self, CborError>;
+}
+
+impl<T> CborDeserialize<T> for T
+where
+    T: for<'de> serde::Deserialize<'de>,
+{
+    fn from_reader<R: std::io::Read>(reader: R) -> Result<Self, CborError> {
+        serde_cbor::de::from_reader(reader).map_err(CborError::from)
+    }
+
+    fn from_slice(slice: &[u8]) -> Result<Self, CborError> {
+        serde_cbor::de::from_slice(slice).map_err(CborError::from)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_indexed::{DeserializeIndexed, SerializeIndexed};
+
+    #[derive(Debug, PartialEq, SerializeIndexed, DeserializeIndexed)]
+    #[serde_indexed(offset = 1)]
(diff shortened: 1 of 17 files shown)


libwebauthn/src/proto/ctap2/cbor/serde.rs
```diff
@@ -0,0 +1,75 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (CborError::SerdeCbor(e1), CborError::SerdeCbor(e2)) => {
+                e1.to_string() == e2.to_string()
+            }
+        }
+    }
+}
+
+pub(crate) trait CborSerialize {
```
msirringhaus: tbh, I'm not much of a fan of having this as a trait. I'd much rather just call `serde_cbor::to_vec(xy)` directly everywhere, as we did before.
If we have the `from()` implementation for `CborError`, handling the errors with `to_vec()?` should even do the error-mapping for us automatically (hopefully). 
Reasons, why I prefer direct calling:
1. It's clear at a glance that this is not a built-in function (the standard lib has many `to_vec()`-functions)
2. Less indirection makes it in general easier to follow
3. Having blanked implementations for all types add up quite a bit of compile time

None of them are super strong reasons against it, but on the other hand, I don't really see a benefit of having it as a trait either. 

Same for trait `CborDeserialize`.

wdyt?

libwebauthn/src/proto/ctap2/cbor/serde.rs
```diff
@@ -0,0 +1,75 @@
...
+        serde_cbor::de::from_reader(reader).map_err(CborError::from)
+    }
+
+    fn from_slice(slice: &[u8]) -> Result<Self, CborError> {
+        serde_cbor::de::from_slice(slice).map_err(CborError::from)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
msirringhaus, line 44: Generally speaking: We should definitely have more cbor-tests. But I'm just as guilty here :sweat_smile: 
So not for this PR, but we should add some more in the future throughout the codebase for underlying parsing-changes like this. Esp. since lot's of the spec depends on the order of fields and so on, which may subtly break if we switch cbor-libs or update them, etc.

msirringhaus (commented):
Added some thoughts

libwebauthn/src/proto/ctap2/cbor/serde.rs
AlfioEmanueleFresta: Discussed offline - there's some value in encapsulating CBOR serialization, but I'll go for two methods rather than traits.

libwebauthn/src/proto/ctap2/cbor/serde.rs
AlfioEmanueleFresta: As the new release of serde-indexed also introduces explicit indexing, I've addressed part of your concern by removing automatic indexing altogether (I'll add copilot as a reviewer to check the indices as it's too mind-numbing of a job)

libwebauthn/src/proto/ctap2/cbor/serde.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    #[test]
+    fn test_deserialize_indexed() {
+        let expected = IndexedStruct { a: 10, b: 20 };
+        let mut map = BTreeMap::new();
+        map.insert(1, 10u8);
+        map.insert(2, 20u8);
+
+        let cbor = to_vec(&map).unwrap();
```
msirringhaus, line 66: Not necessarily for this PR, but I wonder if we could have additionally actual raw (cbor) bytes in these tests to compare the outputs to.

libwebauthn/src/proto/ctap2/cbor/serde.rs
```diff
@@ -0,0 +1,149 @@
...
+pub enum CborError {
+    #[error("serde_cbor serialization error: {0}")]
+    SerdeCbor(#[from] serde_cbor::Error),
+}
+
+impl PartialEq for CborError {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (CborError::SerdeCbor(e1), CborError::SerdeCbor(e2)) => {
+                e1.to_string() == e2.to_string()
```
msirringhaus, line 13: Maybe we could implement `PartialEq` in serde_cbor for this, because this is really ugly :D 
But obviously, not for this PR.

msirringhaus (approved):
Looking good to me! I added two small comments, but good to go as is.

PR #111, opened 2025-05-26
Do not fail on unknown algorithms (fix #66)

### LINKED ISSUE #66: Ignore unknown CTAP response fields

AlfioEmanueleFresta:
Allows parsing unknown algorithm types without failing.

Only ES256 is every used for requests, and it's guaranteed to be supported by the spec, so checking the value is not necessary for now.


libwebauthn/src/proto/ctap1/model.rs
```diff
@@ -12,20 +12,21 @@ use crate::webauthn::CtapError;
 
 #[derive(Debug, Clone, Copy)]
 pub enum Ctap1Transport {
-    BT,
-    BLE,
-    NFC,
-    USB,
+    Bt,
```
msirringhaus: Probably needs rebasing on HEAD?

libwebauthn/src/proto/ctap2/model/credential_protection.rs
```diff
@@ -0,0 +1,22 @@
+use minicbor::{Decode, Encode};
```
msirringhaus: This file is here again :D

libwebauthn/src/proto/ctap2/model.rs
```diff
@@ -242,4 +253,44 @@ mod tests {
...
+            Ctap2CredentialType {
+                algorithm: Ctap2COSEAlgorithmIdentifier::Unknown,
+                public_key_type: Ctap2PublicKeyCredentialType::PublicKey,
+            }
+        );
+        assert!(!credential_type.is_known());
+    }
+
+    #[test]
+    pub fn deerialize_unknown_credential_type() {
```
msirringhaus: typo. `deerialize`

msirringhaus (changes requested):

libwebauthn/src/proto/ctap2/model/credential_protection.rs
AlfioEmanueleFresta: Sorry!

PR #119, opened 2025-06-20
Remove leftover serde_indexed-macros

msirringhaus:
I'm not entirely sure what exactly caused the problem inside serde_indexed, but these macros (particularly the one at `Ctap2GetClientPinRequest`) caused my YubikeyBIO to error out, in the PIN case with `InvalidCborType` when trying to establish a shared secret.


AlfioEmanueleFresta (approved):
Good catch. Sorry I missed these! Thanks for the fix.

PR #120, opened 2025-07-03
Don't block async tasks when reading from HID

iinuwa:
`hid_recv()` is marked as async, but it calls `hid_recv_hidapi()`, which is blocking. This causes the program to hang when the USB device is waiting for user presence.

This wraps the blocking call in `tokio::task::spawn_blocking()` to put in on a separate task pool that won't block other async tasks.

Fixes linux-credentials/linux-webauthn-platform-api#38.


AlfioEmanueleFresta (approved):
Thank you @iinuwa!

PR #125, opened 2025-07-17
Implement discoverable credential 'preferred'

msirringhaus:
Switched away from a context-less `From<MakeCredentialRequest> for Ctap2MakeCredentialRequest` implementation to a dedicated function that can additionally receive the device-info to make that transformation.
Now we also don't need to copy the original request and modify it (which was always a bit icky), but we can do the direct translation to the Ctap2-request.


msirringhaus:
Ugh, the webauthn spec indeed still calls it resident keys. I'll change it.

iinuwa (approved):

iinuwa:
Thanks!

PR #126, opened 2025-07-19
Make Channel async, and return caBLE UX updates (fix #113)

### LINKED ISSUE #113: Send notification for caBLE UI updates
Right now when you create a channel for a caBLE QR code device, you get a reference to a `Receiver<UxUpdate>`, but that only has fields relevant for USB devices. For the UI, we'll need caBLE UI notifications, like Connecting after a BLE advert has been received and Connected after the caBLE tunnel has been established.

We should either:
- add fields to `UxUpdate` to support caBLE UI notifications
- create a transport-specific enum just for caBLE UI notifications.

I think the latter would be cleaner, but would take more shuffling around. I'm fine if we do the former for now and refactor later.

AlfioEmanueleFresta:
Requires #127.

Changes:

* Refactors Hybrid (caBLE) connection stages into 3 reusable components
* Move all 3 connection stages to the channel handle, in addition to existing connection handler
* Operations first wait for a connection to be established, then perform the operation with the operation timeout

Next steps:
* Surface errors

### DIFF
--- a/libwebauthn/src/transport/cable/connection_stages.rs
+++ b/libwebauthn/src/transport/cable/connection_stages.rs
@@ -1,0 +1,322 @@
+use async_trait::async_trait;
+use tokio::sync::{mpsc, watch};
+use tokio_tungstenite::{MaybeTlsStream, WebSocketStream};
+use tracing::{debug, error, instrument};
+
+use super::advertisement::{await_advertisement, DecryptedAdvert};
+use super::channel::{CableUpdate, CableUxUpdate, ConnectionState};
+use super::crypto::{derive, KeyPurpose};
+use super::known_devices::{CableKnownDevice, CableKnownDeviceInfoStore, ClientNonce};
+use super::qr_code_device::CableQrCodeDevice;
+use super::tunnel::{self, CableTunnelConnectionType, TunnelNoiseState};
+use crate::proto::ctap2::cbor::{CborRequest, CborResponse};
+use crate::transport::ble::btleplug::FidoDevice;
+use crate::transport::error::TransportError;
+use crate::webauthn::error::Error;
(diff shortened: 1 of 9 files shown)


libwebauthn/src/transport/cable/connection_stages.rs
```diff
@@ -0,0 +1,322 @@
...
+    let noise_state =
+        tunnel::do_handshake(&mut ws_stream, input.psk, &input.connection_type).await?;
+
+    debug!("Handshake stage completed successfully");
+    ux_sender
+        .send_update(CableUxUpdate::CableUpdate(CableUpdate::Connected))
+        .await;
+
+    ux_sender
+        .set_connection_state(ConnectionState::Connected)
```
msirringhaus, line 299: Do we want to set the state, before sending out the update of the changed state? I'm not fully sure what intricacies there are in terms of race condition possibilities.

msirringhaus (approved):
Looking good from what I can tell. Added one question.

libwebauthn/src/transport/cable/connection_stages.rs
AlfioEmanueleFresta: TL;DR - I think this will be fine as the ConnectionState is only consumed internally, whilst UX updates are only consumed externally.

For each connection state transition:
* The `Connected` state is only used to allow enqueuing CBOR operations, and starting the operation timeout timer. No behaviour that depends on the UX state. 
* The `Terminated` state is only used to fail operations early. If a new operation is queued after disconnection actually occurs, but before the state is transitioned to `Terminated`: the CBOR request will never be dequeued leading to a timeout.

AlfioEmanueleFresta:
Rebased.

Issue #134, opened 2025-07-21
WebAuthn serialization (Web IDL/JSON)


AlfioEmanueleFresta:
### Use cases

* **Authenticators** - Using libwebauthn to communicate with an authenticator
* **Request validation** - Using libwebauthn to validate WebAuthn requests (origin checking) only, then passing the validated JSON request to a WebAuthn IDL-compatible password manager
* **Client processing** - Using libwebauthn to perform WebAuthn client responsibilities (eg. client extension processing) before routing the request to another CTAP2-compatible virtual device (eg. password manager)

### Changes

* [ ] 1. **Deserialization**: Allow WebAuthn operations to be created form WebAuthn JSON, eg. `MakeCredentialRequest::parseJson`. 
  * This should be optional, and it should still be possible to create requests manually.
  * The **origin** of the request should be exposed, allowing the client to perform *origin checking*.
  * The parser should separate **known extensions** from **unknown extensions**:
    * **Known extensions** should continue to be mapped to CTAP2 extension input structures;
    * **Unknown extensions** should be returned as a JSON dictionary, containing the original payload for optional (external) further processing. 
* [ ] 2. **Serialization**: Serialize responses back to WebAuthn JSON

### TODOs

* [ ] Challenge min length validation
* [ ] Allow arbitrary size PRF inputs
* [ ] Update examples to use JSON requests

### Context

* https://github.com/linux-credentials/linux-webauthn-platform-api/issues/26

PR #138, opened 2025-07-27
[WIP] Web IDL support (make credentials); 

AlfioEmanueleFresta:
Not ready to review or merge, opening this for early feedback on the approach.

Next steps:
* Extension parsing from partially parsed JSON (serde_json::Value::Object)
* Get assertion parsing

### DIFF
--- a/libwebauthn/src/ops/webauthn/get_assertion.rs
+++ b/libwebauthn/src/ops/webauthn/get_assertion.rs
@@ -1,3 +1,5 @@
+use super::idl::WebAuthnIDL;
+
 use std::{collections::HashMap, time::Duration};
 
 use serde::{Deserialize, Serialize};
@@ -6,6 +8,11 @@ use tracing::{debug, error, trace};
 
 use crate::{
     fido::AuthenticatorData,
+    ops::webauthn::{
+        create::PublicKeyCredentialCreationOptionsJSON,
+        idl::{FromInnerModel, JsonError},
+        rpid::RelyingPartyId,
+    },
     pin::PinUvAuthProtocol,
     proto::ctap2::{
         Ctap2AttestationStatement, Ctap2GetAssertionResponseExtensions,
--- a/libwebauthn/src/ops/webauthn/make_credential.rs
+++ b/libwebauthn/src/ops/webauthn/make_credential.rs
@@ -7,6 +7,11 @@ use tracing::{debug, instrument, trace};
 
 use crate::{
     fido::AuthenticatorData,
+    ops::webauthn::{
+        create::{AuthenticatorSelectionCriteria, PublicKeyCredentialCreationOptionsJSON},
+        idl::{FromInnerModel, JsonError, WebAuthnIDL},
+        rpid::RelyingPartyId,
+    },
     proto::{
         ctap1::{Ctap1RegisteredKey, Ctap1Version},
         ctap2::{
@@ -149,10 +154,13 @@ impl MakeCredentialsResponseUnsignedExtensions {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Deserialize)]
 pub enum ResidentKeyRequirement {
+    #[serde(rename = "required")]
     Required,
+    #[serde(rename = "preferred")]
     Preferred,
+    #[serde(rename = "discouraged", other)]
     Discouraged,
 }
 
@@ -173,6 +181,68 @@ pub struct MakeCredentialRequest {
     /// extensions
     pub extensions: Option<MakeCredentialsRequestExtensions>,
     pub timeout: Duration,
+}
+
+impl FromInnerModel<PublicKeyCredentialCreationOptionsJSON, MakeCredentialRequestParsingError>
+    for MakeCredentialRequest
+{
+    fn from_inner_model(
+        rpid: &RelyingPartyId,
+        inner: PublicKeyCredentialCreationOptionsJSON,
+    ) -> Result<Self, MakeCredentialRequestParsingError> {
+        let resident_key = if inner
+            .authenticator_selection
+            .as_ref()
+            .and_then(|s| Some(s.require_resident_key))
+            == Some(true)
+        {
+            Some(ResidentKeyRequirement::Required)
+        } else {
+            inner.authenticator_selection.and_then(|s| s.resident_key)
+        };
+
+        let user_verification = inner
+            .authenticator_selection
+            .map_or(UserVerificationRequirement::Discouraged, |s| {
+                s.user_verification
+            });
+
+        let exclude = match inner.exclude_credentials[..] {
+            [] => None,
+            _ => Some(inner.exclude_credentials),
+        };
+
+        let extensions = serde_json::from_value(serde_json::Value::Object(inner.extensions))
+            .map_err(MakeCredentialRequestParsingError::ExtensionError)?;
+
+        Ok(Self {
+            hash: inner.challenge.into(),
+            origin: rpid.as_str().to_owned(),
+            relying_party: inner.rp,
+            user: inner.user,
+            resident_key,
+            user_verification,
+            algorithms: inner.params,
+            exclude,
+            extensions,
+            timeout: Duration::from_secs(inner.timeout),
+        })
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum MakeCredentialRequestParsingError {
+    /// The client must throw an "EncodingError" DOMException.
+    #[error("Invalid JSON format: {0}")]
+    EncodingError(#[from] JsonError),
+
+    #[error("Invalid extension: {0}")]
+    ExtensionError(JsonError),
+}
+
+impl WebAuthnIDL<MakeCredentialRequestParsingError> for MakeCredentialRequest {
+    type Error = MakeCredentialRequestParsingError;
+    type InnerModel = PublicKeyCredentialCreationOptionsJSON;
 }
 
 #[derive(Debug, Default, Clone)]
@@ -315,10 +385,7 @@ impl DowngradableRequest<RegisterRequest> for MakeCredentialRequest {
         }
 
         // Options must not include "rk" set to true.
-        if matches!(
-            self.resident_key,
-            Some(ResidentKeyRequirement::Required)
-        ) {
+        if matches!(self.resident_key, Some(ResidentKeyRequirement::Required)) {
             debug!("Not downgradable: request requires resident key");
             return false;
         }
--- a/libwebauthn/src/ops/webauthn/mod.rs
+++ b/libwebauthn/src/ops/webauthn/mod.rs
@@ -1,5 +1,8 @@
+mod create;
 mod get_assertion;
+pub(crate) mod idl;
 mod make_credential;
+mod rpid;
 
 use super::u2f::{RegisterRequest, SignRequest};
 use crate::webauthn::CtapError;
@@ -17,12 +20,16 @@ pub use make_credential::{
     MakeCredentialResponse, MakeCredentialsRequestExtensions, MakeCredentialsResponseExtensions,
     MakeCredentialsResponseUnsignedExtensions, ResidentKeyRequirement,
 };
+use serde::Deserialize;
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Deserialize)]
 pub enum UserVerificationRequirement {
+    #[serde(rename = "required")]
     Required,
-    Preferred,
+    #[serde(rename = "discouraged")]
     Discouraged,
+    #[serde(rename = "preferred", other)]
+    Preferred,
 }
 
 impl UserVerificationRequirement {
--- a/libwebauthn/Cargo.toml
+++ b/libwebauthn/Cargo.toml
@@ -66,6 +66,7 @@ snow = { version = "0.10", features = ["use-p256"] }
 ctap-types = { version = "0.4.0" }
 btleplug = "0.11.7"
 thiserror = "2.0.12"
+serde_json = "1.0.141"
 
 
 [dev-dependencies]
--- a/libwebauthn/src/ops/webauthn/create.rs
+++ b/libwebauthn/src/ops/webauthn/create.rs
@@ -1,0 +1,67 @@
+use super::idl::Base64UrlString;
+use crate::{
+    ops::webauthn::{ResidentKeyRequirement, UserVerificationRequirement},
+    proto::ctap2::{
+        Ctap2CredentialType, Ctap2PublicKeyCredentialDescriptor, Ctap2PublicKeyCredentialRpEntity,
+        Ctap2PublicKeyCredentialUserEntity,
+    },
+};
+
+use serde::Deserialize;
+use serde_json::{Map as JsonMap, Value as JsonValue};
+
+/**
+ * https://www.w3.org/TR/webauthn-3/#sctn-parseCreationOptionsFromJSON
+ */
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct PublicKeyCredentialDescriptorJSON {
+    pub r#type: String,
+    pub id: Base64UrlString,
+    pub transports: Vec<String>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct PublicKeyCredentialParameters {
+    pub r#type: String,
+    pub alg: i32,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct AuthenticatorSelectionCriteria {
+    #[serde(rename = "authenticatorAttachment")]
+    pub authenticator_attachment: Option<String>,
+    #[serde(rename = "residentKey")]
+    pub resident_key: Option<ResidentKeyRequirement>,
+    #[serde(rename = "requireResidentKey")]
+    #[serde(default)]
+    pub require_resident_key: bool,
+    #[serde(rename = "userVerification")]
+    #[serde(default = "default_user_verification")]
+    pub user_verification: UserVerificationRequirement,
+}
+
+fn default_user_verification() -> UserVerificationRequirement {
+    UserVerificationRequirement::Preferred
+}
+
+type JsonObject = JsonMap<String, JsonValue>;
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct PublicKeyCredentialCreationOptionsJSON {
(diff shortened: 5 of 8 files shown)


AlfioEmanueleFresta:
Note to self, the failures are due to Base64UrlString being used in CTAP models, which are serialized as a Base64 string rather than URL with serde_cbor. The next step is to separate these models into two, a WebAuthn JSON model using Base64UrlString, and an equivalent CTAP model using ByteBuf. Must also review all usages of Base64UrlString.

AlfioEmanueleFresta:
Marking this as ready for review as this is getting large, keeping track of further work in https://github.com/linux-credentials/libwebauthn/issues/134.

libwebauthn/examples/prf_test.rs
```diff
@@ -126,10 +126,11 @@ pub async fn main() -> Result<(), Box<dyn Error>> {
         });
 
         let eval_by_credential = HashMap::new();
-        let hmac_or_prf = GetAssertionHmacOrPrfInput::Prf {
+        let hmac_or_prf: GetAssertionHmacOrPrfInput = GetAssertionHmacOrPrfInput::Prf(PrfInput {
```
msirringhaus: Is the type specifier `let hmac_or_prf: GetAssertionHmacOrPrfInput` needed? I think it could be removed.

libwebauthn/examples/webauthn_prf_hid.rs
```diff
@@ -148,10 +148,10 @@ pub async fn main() -> Result<(), Box<dyn Error>> {
                 second: None,
             },
         );
-        let hmac_or_prf = GetAssertionHmacOrPrfInput::Prf {
+        let hmac_or_prf: GetAssertionHmacOrPrfInput = GetAssertionHmacOrPrfInput::Prf(PrfInput {
```
msirringhaus: Same here and below

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -63,17 +74,17 @@ impl MakeCredentialsResponseUnsignedExtensions {
         let mut prf = None;
         if let Some(signed_extensions) = signed_extensions {
             (hmac_create_secret, prf) = if let Some(incoming_ext) = &request.extensions {
-                match &incoming_ext.hmac_or_prf {
-                    MakeCredentialHmacOrPrfInput::None => (None, None),
-                    MakeCredentialHmacOrPrfInput::HmacGetSecret => {
-                        (signed_extensions.hmac_secret, None)
-                    }
-                    MakeCredentialHmacOrPrfInput::Prf => (
+                if let Some(_hmac_create_secret) = incoming_ext.hmac_create_secret {
```
msirringhaus: Using `if {} else if {}` here means you can either have HMAC or PRF, and if you request both, HMAC will override PRF. Is that intended?

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -20,6 +29,8 @@ use crate::{
 
 use super::{DowngradableRequest, RegisterRequest, UserVerificationRequirement};
 
+pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(60);
```
msirringhaus: Should we use the defintion from `libwebauthn/src/ops/webauthn/timeout.rs` here? (Same in the other files where this is defined again)

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -14,35 +25,190 @@ use crate::{
...
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct GetAssertionRequest {
     pub relying_party_id: String,
     pub hash: Vec<u8>,
     pub allow: Vec<Ctap2PublicKeyCredentialDescriptor>,
-    pub extensions: Option<GetAssertionRequestExtensions>,
+    pub extensions: GetAssertionRequestExtensions,
```
msirringhaus: I confess, I'm currently a bit lost in the details of this huge PR, but if I recall correctly, we used to use the `Option<>` here to decide if we have to include an extensions-map in the output. Some portals like demo.yubico.com are rather strict on what they expect, e.g. if they request extensions, they expect `extensions` in the response, even if the map is empty. And they do not want to see that empty map, if they didn't request one.
Is this still possible with this change?

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -14,35 +25,190 @@ use crate::{
     webauthn::CtapError,
 };
 
-use super::{DowngradableRequest, SignRequest, UserVerificationRequirement};
+use super::{DowngradableRequest, RelyingPartyId, SignRequest, UserVerificationRequirement};
 
-#[derive(Debug, Default, Clone, Serialize)]
+pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(60);
```
msirringhaus: Same question here as elsewhere: Should this be redefined here or imported?

msirringhaus (commented):
Looking good overall, I think. I haven't had time for a full deep dive, though. I only have some preliminary questions.

libwebauthn/src/ops/webauthn/get_assertion.rs
AlfioEmanueleFresta: Thanks for spotting this, addressing it so we can represent both states. Also adding a test for both cases (`extensions: {}` and no `etensions`).

libwebauthn/src/ops/webauthn/make_credential.rs
AlfioEmanueleFresta: Hmm, I think you're right. Whilst I wouldn't expect a browser to request this, I'll change it so that if your makeCredential includes both HMAC & PRF, you will get both responses back.

libwebauthn/examples/webauthn_prf_hid.rs
msirringhaus: I don't think we need the double type marker here
```suggestion
        let hmac_or_prf = GetAssertionHmacOrPrfInput::Prf(PrfInput {
```

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -34,15 +46,170 @@ pub struct GetAssertionRequest {
...
+                .map(|extensions_opt| GetAssertionRequestExtensions {
+                    cred_blob: extensions_opt.cred_blob.unwrap_or(false),
+                    large_blob: extensions_opt
+                        .large_blob
+                        .clone()
+                        .map(Option::<GetAssertionLargeBlobExtension>::try_from)
+                        .transpose()
+                        .ok()
+                        .flatten()
+                        .flatten(),
```
msirringhaus: Can this be done a little bit simpler? Feels like transpose + 2x flatten points to an overly complicated data structure. E.g. does the `try_from()` function need to be implemented for `Option<GetAssertionLargeBlobExtension>` instead of `GetAssertionLargeBlobExtension` or `&GetAssertionLargeBlobExtension`? Then we could maybe do `and_then()` instead of `map()` here.

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -34,15 +46,170 @@ pub struct GetAssertionRequest {
...
+                                        "extensions.prf.eval_by_credential[i].second".to_string(),
+                                        s.as_slice().len(),
+                                    )
+                                })?),
+                                None => None,
+                            },
+                        },
+                    ))
+                })
+                .collect::<Result<HashMap<String, PRFValue>, GetAssertionRequestParsingError>>()?,
```
msirringhaus, line 201: This can probably be shortened by omitting at least the error type
```suggestion
                .collect::<Result<HashMap<String, PRFValue>, _>>()?,
```
The compiler should be able to figure it out (same for String and PRFValue, but those are not overly long)

libwebauthn/src/ops/webauthn/idl/create.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use serde::Deserialize;
+
+/**
+ * https://www.w3.org/TR/webauthn-3/#sctn-parseCreationOptionsFromJSON
+ */
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct AuthenticatorSelectionCriteria {
+    #[serde(rename = "authenticatorAttachment")]
```
msirringhaus: I think you can use `#[serde(rename_all = "camelCase")]` here instead of renaming each field individually.

libwebauthn/src/ops/webauthn/idl/create.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub authenticator_attachment: Option<String>,
+    #[serde(rename = "residentKey")]
+    pub resident_key: Option<ResidentKeyRequirement>,
+    #[serde(rename = "requireResidentKey")]
+    #[serde(default)]
+    pub require_resident_key: bool,
+    #[serde(rename = "userVerification")]
+    #[serde(default = "default_user_verification")]
```
msirringhaus, line 24: We could think about deriving `Default` for `UserVerificationRequirement` and marking `Preferred` with `#[default]` and remove `default_user_verification()`. But maybe we don't want a default impl for that. Would be nice if serde would finally implement the ability to add default values here.

libwebauthn/src/ops/webauthn/idl/create.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct PublicKeyCredentialUserEntity {
+    pub id: Base64UrlString,
+    pub name: String,
+    #[serde(rename = "displayName")]
+    pub display_name: String,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct PublicKeyCredentialCreationOptionsJSON {
```
msirringhaus, line 42: I'd also  go for `rename_all` here

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -149,14 +160,17 @@ impl MakeCredentialsResponseUnsignedExtensions {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Deserialize, PartialEq)]
```
msirringhaus, line 163: This could also be done with `rename_all`

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -254,13 +329,20 @@ pub struct CredentialPropsExtension {
...
+#[derive(Debug, Default, Clone, Deserialize, PartialEq)]
+pub struct MakeCredentialLargeBlobExtensionInput {
+    pub support: MakeCredentialLargeBlobExtension,
+}
+
+#[derive(Debug, Default, Clone, Copy, Deserialize, Serialize, PartialEq, Eq)]
 pub enum MakeCredentialLargeBlobExtension {
-    #[default]
-    None,
+    #[serde(rename = "preferred")]
```
msirringhaus: `rename_all`

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -269,14 +351,22 @@ pub struct MakeCredentialLargeBlobExtensionOutput {
     pub supported: Option<bool>,
 }
 
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Default, Clone, Deserialize, PartialEq)]
```
msirringhaus, line 354: `rename_all`

libwebauthn/src/ops/webauthn/mod.rs
```diff
@@ -8,21 +11,32 @@ pub use get_assertion::{
...
+use serde::Deserialize;
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Deserialize, PartialEq)]
+pub enum Operation {
+    MakeCredential,
+    GetAssertion,
+}
+
+#[derive(Debug, Clone, Copy, Deserialize, PartialEq)]
```
msirringhaus: `rename_all`

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -175,22 +189,81 @@ pub struct MakeCredentialRequest {
...
+
+impl WebAuthnIDL<MakeCredentialRequestParsingError> for MakeCredentialRequest {
+    type Error = MakeCredentialRequestParsingError;
+    type InnerModel = PublicKeyCredentialCreationOptionsJSON;
+}
+
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+pub struct MakeCredentialPrfInput {
+    #[serde(rename = "eval")]
+    pub _eval: Option<JsonValue>,
```
msirringhaus, line 269: Can we have a comment here, why it's `_eval`, aka not used at the moment?

libwebauthn/src/ops/webauthn/client_data.rs
```diff
@@ -0,0 +1,36 @@
...
+use serde::Deserialize;
+use sha2::{Digest, Sha256};
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct ClientData {
+    pub operation: Operation,
+    pub challenge: Vec<u8>,
+    pub origin: String,
+    #[serde(rename = "crossOrigin")]
+    pub cross_origin: Option<bool>,
```
iinuwa, line 12: Can we add `topOrigin` to this as well?
https://www.w3.org/TR/webauthn-3/#dom-collectedclientdata-toporigin

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -227,3 +429,162 @@ impl DowngradableRequest<Vec<SignRequest>> for GetAssertionRequest {
...
+    }
+
+    #[test]
+    fn test_request_from_json_ignore_missing_rp_id() {
+        let rpid = RelyingPartyId::try_from("example.org").unwrap();
+        let req_json = json_field_rm(REQUEST_BASE_JSON, "rpId");
+
+        let req: GetAssertionRequest = GetAssertionRequest::from_json(&rpid, &req_json).unwrap();
+        assert_eq!(req, request_base());
+    }
```
iinuwa, line 510: This is more of a note for me/credentialsd: even though it's valid to omit the `rp.id`/`rpId` field because the client can infer the RP ID from the effective domain of the origin, I think that credentialsd should require setting the RP ID for simplicity.

libwebauthn/src/ops/webauthn/idl/base64url.rs
```diff
@@ -0,0 +1,67 @@
...
+        S: serde::Serializer,
+    {
+        let encoded = base64_url::encode(&self.0);
+        serializer.serialize_str(&encoded)
+    }
+}
+
+impl Into<Vec<u8>> for Base64UrlString {
+    fn into(self) -> Vec<u8> {
+        self.0
```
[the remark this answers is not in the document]
iinuwa: Agreed

libwebauthn/src/ops/webauthn/idl/get.rs
```diff
@@ -0,0 +1,81 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub transports: Option<Vec<Ctap2Transport>>,
+}
+
+impl Into<Ctap2PublicKeyCredentialDescriptor> for PublicKeyCredentialDescriptorJSON {
+    fn into(self) -> Ctap2PublicKeyCredentialDescriptor {
+        Ctap2PublicKeyCredentialDescriptor {
+            r#type: self.r#type,
+            id: ByteBuf::from(self.id),
+            transports: self.transports,
```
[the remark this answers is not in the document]
iinuwa: Agreed

libwebauthn/src/ops/webauthn/idl/rpid.rs
```diff
@@ -0,0 +1,49 @@
...
+    type Target = str;
+
+    fn deref(&self) -> &str {
+        &self.0
+    }
+}
+
+impl Into<String> for RelyingPartyId {
+    fn into(self) -> String {
+        self.0
```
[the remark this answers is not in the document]
iinuwa: Agreed

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -75,11 +277,11 @@ pub struct GetAssertionLargeBlobExtensionOutput {
...
 }
 
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Default, Clone, PartialEq)]
 pub struct GetAssertionRequestExtensions {
-    pub cred_blob: Option<bool>,
-    pub hmac_or_prf: GetAssertionHmacOrPrfInput,
-    pub large_blob: GetAssertionLargeBlobExtension,
+    pub cred_blob: bool,
+    pub hmac_or_prf: Option<GetAssertionHmacOrPrfInput>,
```
iinuwa, line 282: Since this is in a WebAuthn-specific module, can we make this just `prf: PrfInput`?  The relationship between WebAuthn PRF and HMAC secret is pretty confusing, and it'd be more confusing to make it seem like this type could accept HMAC input when it's really only PRF at WebAuthn level.

If it needs to be wrapped in `GetAssertionHmacOrPrfInput` later in the API, that's fine.

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -269,14 +351,22 @@ pub struct MakeCredentialLargeBlobExtensionOutput {
...
-    pub large_blob: MakeCredentialLargeBlobExtension,
+    #[serde(rename = "credBlob")]
+    pub cred_blob: Option<Base64UrlString>,
+    #[serde(rename = "largeBlob")]
+    pub large_blob: Option<MakeCredentialLargeBlobExtensionInput>,
+    #[serde(rename = "minPinLength")]
     pub min_pin_length: Option<bool>,
-    pub hmac_or_prf: MakeCredentialHmacOrPrfInput,
+    #[serde(rename = "hmacCreateSecret")]
+    pub hmac_create_secret: Option<bool>,
```
iinuwa: Are there any RPs that are using this? At the WebAuthn level, there's only PRF, and I'd like to keep it that way if possible.

libwebauthn/src/ops/webauthn/get_assertion.rs
AlfioEmanueleFresta: Sadly the compiler won't accept this

PR #136, opened 2025-07-27
Support attested_credential and extensions both be present in AuthenticatorData

jo-bitsch:
As described in #135, AuthenticatorData can contain both an attested_credential, as well as extensions when making a credential. However, from_reader checks that there is no trailing data in the cbor of the public key by calling deserializer.end(). Therefore, we need to use the manual interface to deserialize.

Fixes #135.

PR #139, opened 2025-08-03
extended the tests for AuthData deserialization

jo-bitsch:
As suggested by @AlfioEmanueleFresta in #136 in https://github.com/linux-credentials/libwebauthn/pull/136#pullrequestreview-3059589590 added a test for deserializing AuthData.

To be able to test this, I had to pull out the AuthenticatorDataVisitor struct to make the visit_bytes function available, I understand that this exposes the implementation details a bit, so might not be the best solution here. I'm open for suggestions.

### DIFF
--- a/libwebauthn/src/fido.rs
+++ b/libwebauthn/src/fido.rs
@@ -229,10 +229,15 @@ impl<'de, T: DeserializeOwned> Deserialize<'de> for AuthenticatorData<T> {
 
                 let extensions: Option<T> =
                     if flags.contains(AuthenticatorDataFlags::EXTENSION_DATA) {
-                        cbor::from_reader(&mut cursor).map_err(DesError::custom)?
+                        cbor::from_cursor(&mut cursor).map_err(DesError::custom)?
                     } else {
                         Default::default()
                     };
+
+                // Check if we have trailing data
+                if !&data[cursor.position() as usize..].is_empty() {
+                    return Err(DesError::invalid_length(data.len(), &"trailing data"));
+                }
 
                 Ok(AuthenticatorData {
                     rp_id_hash,
@@ -251,6 +256,9 @@ impl<'de, T: DeserializeOwned> Deserialize<'de> for AuthenticatorData<T> {
 #[cfg(test)]
 mod tests {
     use cosey::{Bytes, Ed25519PublicKey};
+    use serde_bytes::ByteBuf;
+
+    use crate::proto::ctap2::cbor;
 
     use super::{AttestedCredentialData, AuthenticatorData, AuthenticatorDataFlags};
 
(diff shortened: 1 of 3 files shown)


msirringhaus:
I think you should be able to leave `struct AuthenticatorDataVisitor<T>` untouched.
The goal is basically to create an `AuthenticatorData`-object, serialize it via `auth_data.to_response_bytes()`, inspect the individual byte-chunks for validity, and then as a new test, deserialize it again to an `AuthenticatorData`-object, and verify that the original object and the serialized and then deserialized object are identical, correct?

You don't have to expose the `AuthenticatorDataVisitor<T>`-struct to deserialize the bytes, you should be able to simply put them in `cbor::from_slice(..)`.
So something like 
```
        let auth_data_reparsed: AuthenticatorData<String> =
            cbor::from_slice(&webauthn_auth_data).unwrap();
        assert_eq!(extensions, auth_data_reparsed.extensions);
        // maybe assert_eq all other members of auth_data_reparsed as well
```
in the test function. (Note quite sure if all the types are correct here, but as a rough direction.)

jo-bitsch:
@msirringhaus I adjusted the PR to use the deserializer as is. There was the minor complication that `cbor::from_slice()` requires a cbor formated input, so for the test, I wrapped the bytes again using `cbor::to_vec(&ByteBuf::from(webauthn_auth_data))`.

libwebauthn/src/fido.rs
```diff
@@ -312,6 +323,13 @@ mod tests {
             &credential_id,
             &webauthn_auth_data[55..55 + &credential_id.len()]
         );
-        assert_eq!(cose_bytes, &webauthn_auth_data[55 + credential_id.len()..]);
+        let extensions_bytes = cbor::to_vec(&extensions).unwrap();
+        assert_eq!(cose_bytes, &webauthn_auth_data[55 + credential_id.len()..webauthn_auth_data.len() - extensions_bytes.len()]);
+
+        let authdata_wrapped = cbor::to_vec(&ByteBuf::from(webauthn_auth_data)).unwrap();
```
msirringhaus, line 341: Can you add a comment here, basically stating shortly what we are trying to do here? Something like `// Deserialize the serialized data again and verify it's the same as the original`, but maybe better worded :sweat_smile:

libwebauthn/src/fido.rs
```diff
@@ -312,6 +323,13 @@ mod tests {
...
+        let auth_data_reparsed: AuthenticatorData<String> =
+            cbor::from_slice(authdata_wrapped.as_slice()).unwrap();
+        assert_eq!(extensions, auth_data_reparsed.extensions);
```
msirringhaus: As we already have everything in place here, can you add `assert_eq!(..)` for all the other fields (rp_id_hash, flags, ...) of original vs. reparsed as well?

msirringhaus (changes requested):
Thanks! This looks pretty good now! I was indeed afraid that you'd have to do a little cbor format dance. Hope, it didn't cost you too much time figuring it out. 

I added a few minor requests, that would be nice to have, but these should be very quick and simple.

PR #140, opened 2025-08-11
Fix PadError when doing HMAC calculation

msirringhaus:
This `to_vec()` was accidentally rewritten to `cbor::to_vec()` in 8ac81b80, causing the result to be not a multiple of the AES blocksize.


iinuwa (approved):
Thanks for catching this!

PR #143, opened 2025-08-21
Switch to fido-authenticator from Nitrokey for virtual HID device

msirringhaus:
A few things happening here. Please excuse the mess of a commit-history. Best to ignore it and squash :sweat_smile: 

- Removed solo-key and switched to fido-authenticator (for both ctap1 and ctap2)
- Removed top-level Cargo-workspace, as we only have one crate now
- Removed the basically unused `hid-device-tests` feature 
- Adjusted github CI to also run the webauthn_hid example

Implementation details:
- The trussed-ecosystem is a bit of a mess, so very special versions are required for this to work. So we have to patch crates.io in the Cargo.toml file. This is based on the way fido-authenticator is using it in it's testsuite.
- Introduced a new `VirtHidDevice` that is only a stub without the feature flag to avoid the cfg-checks all over the place, and have them bundled in one file, instead.
- `libwebauthn/src/transport/virt/pipe.rs` was copied from fido-authenticator, which in turn copied it from trussed. I kept the original pointer to trussed in the header.
- `libwebauthn/src/transport/virt/device.rs` was copied from fido-authenticator and slightly modified (see below). Both files are MIT, so including it with source-links should be fine, I think?
- Used certificate and key are copied from nitrokey-3-firmware (also MIT). We could potentially look into generating them ourselves, instead. Maybe even during compilation, to avoid inclusion of binary files in git.
- fido-authenticator is using it with a RAM-storage implementation, which is lightweight, but not suitable for our usage, as we want to run our examples, that chain multiple commands and expect state to be kept. Due to the code structure of trussed, RAM-storage would have been wiped with each command, so I opted for a simple temporary filesystem storage instead, which gets removed, when the virtual device is dropped.

Not yet done (probably something for future PRs):
- Make all examples work with the virtual device. Most of the examples don't dynamically adjust their command usage (e.g. extensions) based on the capability of the device, but do a battery of commands hardcoded. Some of those fail due to lack of support in the virtual device. 
- Create an extensive testsuite that runs a lot of variations to cover as much of the feature set as possible.


libwebauthn/src/transport/hid/device.rs
```diff
@@ -1,34 +1,25 @@
...
     pub backend: HidBackendDevice,
 }
 
-#[derive(Debug)]
-// SoloVirtualKey is not clone-able, but in test-mode we don't care
-#[cfg_attr(not(feature = "virtual-hid-device"), derive(Clone))]
+#[derive(Debug, Clone)]
 pub enum HidBackendDevice {
     HidApiDevice(DeviceInfo),
-    #[cfg(feature = "virtual-hid-device")]
```
AlfioEmanueleFresta, line 30: Would it be easier to feature-gate the code in mod.rs (`pub use`), and here?

It would remove the need for all the gating within `virt`, and the mock device which doesn't provide much value.

libwebauthn/src/transport/hid/device.rs
```diff
@@ -102,23 +91,3 @@ impl<'d> Device<'d, Hid, HidChannel<'d>> for HidDevice {
     //     channel.supported_protocols().await
     // }
 }
-
-#[cfg(test)]
-mod tests {
-
-    #[cfg(feature = "hid-device-tests")]
```
AlfioEmanueleFresta, line 109: I agree with removing this feature as we only had one test, but I would love to have some E2E tests run automatically on `cargo test`.

I think we would want tests to be as comprehensive as possible, whilst examples would merely need to be readable/understandable - so they serve different purposes.

libwebauthn/src/transport/virt/device.rs
```diff
@@ -0,0 +1,239 @@
...
+use trussed::{
+    backend::BackendId,
+    platform::Platform as _,
+    store::Store as _,
+    virt::{self, StorageConfig, StoreConfig},
+};
+use trussed_staging::virt::{BackendIds, Client, Dispatcher};
+
+// see: https://github.com/Nitrokey/nitrokey-3-firmware/tree/main/utils/test-certificates/fido
+const ATTESTATION_CERT: &[u8] = include_bytes!("../../../../data/fido-cert.der");
```
AlfioEmanueleFresta: Should this be within `libwebauthn`? Not sure if the crates.io build would succeed otherwise.

Ideally, can we use the cargo environment variable here?

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

AlfioEmanueleFresta (changes requested):
Thank you very much @msirringhaus, this is great. 

This LGTM, added two comments on feature gating and test key material paths.

libwebauthn/Cargo.toml
```diff
@@ -67,6 +66,19 @@ ctap-types = { version = "0.4.0" }
 btleplug = "0.11.7"
 thiserror = "2.0.12"
 
+# For feature: "virtual-hid-device"
+ctaphid-dispatch = { version = "0.3", optional = true }
+ctaphid = { version = "0.3.1", default-features = false, optional = true }
+fido-authenticator = { git = "https://github.com/Nitrokey/fido-authenticator.git",tag = "v0.1.1-nitrokey.27", features = ["dispatch"], optional = true }
```
AlfioEmanueleFresta: Can we use a commit hash?

libwebauthn/src/transport/hid/device.rs
msirringhaus: Then we also have to feature-gate every `match` statement for `HidBackendDevice`, which is why I removed the gate here and opted for the mock implementation instead.
Felt cleaner (aka less places needed, where we need the feature-gate) and somewhat less annoying to work with for me, but I have no strong opinions here. Both options would be fine with me.

libwebauthn/src/transport/hid/device.rs
msirringhaus: Yes, I plan to do that

libwebauthn/src/transport/virt/device.rs
msirringhaus: It's now compiled in only for `cfg(test)`, so hopefully it will be ok for crates.io. 
But I really want to get rid of them completely and generate them live, if possible, in the future. So if the current PR doesn't work, I'll have to do that first.

libwebauthn/src/transport/hid/device.rs
msirringhaus: Now examples are not using virtual devices at all (because cargo won't allow me to do that, sadly), and all E2E-tests are run upon `cargo test`. Also in the CI.

libwebauthn/src/transport/hid/device.rs
msirringhaus: Now that I have it gated behind `cfg(test)`, I reverted to the old version you preferred (hiding the enum-variant and all `match`-places, instead of having a mock virtual device), since my version made way less sense now.

libwebauthn/src/transport/virt/device.rs
```diff
@@ -0,0 +1,239 @@
...
+use trussed::{
+    backend::BackendId,
+    platform::Platform as _,
+    store::Store as _,
+    virt::{self, StorageConfig, StoreConfig},
+};
+use trussed_staging::virt::{BackendIds, Client, Dispatcher};
+
+// see: https://github.com/Nitrokey/nitrokey-3-firmware/tree/main/utils/test-certificates/fido
+const ATTESTATION_CERT: &[u8] = include_bytes!("./data/fido-cert.der");
```
AlfioEmanueleFresta, line 44: Neat! I didn't know this existed.

PR #146, opened 2025-08-28
Allow for shared secret to be created even if no PIN is set on the device

msirringhaus:
This is to support HMAC/PRF on such devices, which currently only worked for devices with PIN or UV set (i.e. devices that could negotiate a `pinUvAuthToken`).

Things of note:
- The requests now have a `needs_shared_secret()` function that is only returning true for GetAssertion at the moment. I have not checked all the other commands in detail, if they could potentially also require this. If so, it would be easy to fill in (e.g. for Ctap2.2 where MakeCredential could also do HMAC).
- In the `AuthTokenData`-store, `pinUvAuthToken` (and associated permissions) are now optional. Luckily all this is hidden from the rest of the library, as we already had an accessor-function for this that returned an Option.
- The `user_verification_helper()` function is getting a bit more complex, sadly. But I hope with all my comments, it's not too bad. Making the naming of `Ctap2UserVerificationOperation`-variations more explicit should also help.

### DIFF
--- a/libwebauthn/src/proto/ctap2/model/get_info.rs
+++ b/libwebauthn/src/proto/ctap2/model/get_info.rs
@@ -6,7 +6,7 @@ use tracing::debug;
 
 use super::{Ctap2CredentialType, Ctap2UserVerificationOperation};
 
-#[derive(Debug, Clone, DeserializeIndexed)]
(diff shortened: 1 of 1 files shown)


AlfioEmanueleFresta:
I have no 2.0 hardware to test the behaviour with, but the code looks good to me.

Thank you!

PR #149, opened 2025-09-11
Add more UV-negotiation tests

msirringhaus:
This PR adds another layer of testing opportunities, by introducing a `TestChannel` that can be pre-filled with expected requests and associated responses it should give.

I have used this to test `user_verification(..)` in various states of complexity, and discovered a few bugs along the way.

Notable changes to make this possible:
- `PinUvAuthProtocol`s get a deterministic private key, to 'predict' crypto results (for testing only!)
- Some (the ones I needed) Ctap2-response structs got a `#[cfg_attr(test, derive(SerializeIndexed))]` to serialize them to CBOR, before sending it back as if it were a real message from a device. This should be easy to extend to other response structs, once they are needed.
- Removed a second, subsequent `GetInfo`-call, by simply reusing the result of the first call in `user_verification(..)`
- Fixed some corner cases in `user_verification(..)`, mostly around the "shared secret only"-scenarios
- Renamed `ClientPinOnlyForSharedSecret` to `OnlyForSharedSecret` as this is not bound to `ClientPin` anymore
- Added loads of tests. Still not exhaustive, though. More should be written around full ceremony scenarios (I'll open tickets for some of those, once this lands).

### DIFF
--- a/libwebauthn/src/pin.rs
+++ b/libwebauthn/src/pin.rs
@@ -10,7 +10,7 @@ use p256::{
     ecdh::EphemeralSecret, elliptic_curve::sec1::FromEncodedPoint, EncodedPoint,
     PublicKey as P256PublicKey,
 };
-use rand::{rngs::OsRng, thread_rng, Rng};
+use rand::{rngs::OsRng, thread_rng, Rng, SeedableRng};
 use sha2::{Digest, Sha256};
 use tracing::{error, instrument, warn};
 use x509_parser::nom::AsBytes;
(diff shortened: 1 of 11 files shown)


libwebauthn/src/pin.rs
```diff
@@ -97,7 +97,14 @@ pub struct PinUvAuthProtocolOne {
 
 impl PinUvAuthProtocolOne {
     pub fn new() -> Self {
-        let private_key = EphemeralSecret::random(&mut OsRng);
+        let private_key = if cfg!(test) {
+            // For testing only!!
+            // We need a deterministic, seedable RNG to be able to "predict" crypto-operations
+            let mut rng = rand::rngs::StdRng::seed_from_u64(42);
```
AlfioEmanueleFresta, line 103: Nit: In the spirit of minimising these behaviour divergences, could we define a constant RNG - with different values in test/non-test - and re-use it?

msirringhaus: You mean: have a single pseudo RNG with different (random?) seeds for test/non-test? 
In principle, yes, but I'm not totally sure what advantage that would gain us, but it would be - arguably - 'less secure' in the non-test case, which currently goes through `getrandom()` that also utilizes real hardware entropy (on reasonably modern kernels).

PR #150, opened 2025-09-12
Switch virt device to RAM only and use message passing to allow for state to be kept across device lifetime

msirringhaus:
Note: Needs rebasing after #149 lands, as this built on top of it.

### Commit 1:
Since the callback structure of nitrokeys/trusseds firmware makes it hard to have a 'normal' device, that can handle more than one consecutive requests without losing state, and after banging my head against this for quite some time, I opted for a rather blunt, but hopefully effective (enough) solution: Simple message passing. 

Now the callback-function is started in its own thread and requests+responses are passed back and forth via normal channels. 
I also removed the filesystem storage and went back to RAM storage only. This way, it should be possible to have multiple devices in parallel without any interference. As long as the device object lives, state is kept and things like "establishing shared secret, getting pinUvAuthToken, then doing the request" work, as the device is not reset between each request.

I added a rudimentary test-battery for PRF, demonstrating that this can now be done.

I also added a few more dev-dependencies to allow for better usage/debugging of this virtual device. With these, it is possible to show logging from the virtual fido-authenticator in the end2end-tests.

### Commit 2:
Renaming `TestChannel` to `MockChannel` (which I think explains better what it is) and moving the file out of `src/tests/`. I realized this would be very confusing, as in there are only tests that use the virtual device and not the `MockChannel`.
So now we have two testing transports under `src/transports/`, namely `virt` and `mock`, which hopefully is a bit more self-explanatory.

### DIFF
--- a/libwebauthn/src/transport/virt/mod.rs
+++ b/libwebauthn/src/transport/virt/mod.rs
@@ -5,89 +5,123 @@ use super::hid::framing::HidCommand;
 use super::hid::framing::HidMessage;
 use crate::webauthn::Error;
 use num_enum::TryFromPrimitive;
-use std::sync::Arc;
+use std::sync::mpsc::{Receiver, Sender};
+use std::thread;
+use std::thread::JoinHandle;
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub(crate) struct VirtHidDevice {
-    storage_dir: Arc<tempfile::TempDir>,
-    response: Option<HidMessage>,
+    req_tx: Sender<HidMessage>,
+    resp_rx: Receiver<HidMessage>,
+    _device_thread: JoinHandle<()>,
+}
+
+impl Drop for VirtHidDevice {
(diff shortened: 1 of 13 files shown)


libwebauthn/src/transport/virt/mod.rs
```diff
@@ -5,89 +5,123 @@ use super::hid::framing::HidCommand;
...
+
+        // Due to the callback-structure of trussed, we simply can't call
+        // device::run_ctaphid() multiple times, as that would mean a new
+        // device initialization for each call (and rendering every shared
+        // secret invalid).
+        // So we let it run only once, but in a seperate thread, and simply
+        // pass messages back and forth.
+        // This way, the device works across multiple request calls,
+        // as long as the device-object lives. Everything is wiped on dropping
+        // the device object.
```
AlfioEmanueleFresta, line 41: Neat!

AlfioEmanueleFresta (approved):

PR #151, opened 2025-09-26
Implement pin protocol tests

msirringhaus:
Based on #150 (needs rebasing after that will land): Add a way to force the pin protocol supported by the platform, in order to test both protocols for various actions, such as setting and changing PINs and doing PRF.


AlfioEmanueleFresta (approved):
Thank you!

PR #152, opened 2025-10-01
Implement NFC

msirringhaus:
Very heavily based on #105, with the following additions: 

- Fleshed out the APDU-calls
- Added `select_fido2()` to discover both U2F and FIDO2-only devices
- Made `NfcDevice` derive `Clone` and other minor tweaks, so we can easily use it in `credentialsd`
- Added the build with both NFC-features to the github CI (including the installation of libnfc et al.). I kept this separate from the original, un-featured build, so we can be sure the former works even without libnfc installed.

Open questions:
- Biggest question to me is, how to handle the multiple backends and multiple devices. Since we can't run the "blink and wait for user presence"-scheme with NFC as we do with USB, I'm leaning towards simply always returning the very first found NFC-device instead of a list of found devices and use that. With this, we could get rid of the duplication-problem. I'm not sure many setups would include more than one NFC-reader (with different devices on there).
- The cancel-handle is currently unused, and I'm not sure if we can use it anywhere at all.
- No testing yet. Not sure how easy that would be to add.
- Given that `credentialsd` uses a poll-mechanism to find USB-devices (simply looping over `list_devices()`) as well, I did not investigate to get a blocking device discovery.
- For testing, I hacked my local credentialsd-repo by simply replacing `hid` with `nfc` in the USB-handler. Thus, I needed a dummy `blink_and_wait_for_user_presence()`-implementation. This should probably be removed, esp. if we go for the "return first device"-scheme mentioned above. But doing that, I was able to successfully use NFC-devices with Firefox.


libwebauthn/examples/u2f_hid.rs
```diff
@@ -33,6 +36,14 @@ async fn handle_updates(mut state_recv: Receiver<UvUpdate>) {
 pub async fn main() -> Result<(), Box<dyn Error>> {
     setup_logging();
 
+    #[cfg(feature = "nfc")]
```
AlfioEmanueleFresta: Could we have NFC specific tests? A functional HID example failing if I _enable_ a feature feels a bit of an anti-pattern. (I also personally use `--all-targets` quite a lot, and this would break that flow.)

libwebauthn/src/transport/nfc/channel.rs
```diff
@@ -0,0 +1,335 @@
...
+    async fn supported_protocols(&self) -> Result<SupportedProtocols, Error> {
+        Ok(self.supported)
+    }
+
+    async fn status(&self) -> ChannelStatus {
+        self.status
+    }
+
+    async fn close(&mut self) {
+        todo!("close")
```
AlfioEmanueleFresta: Leftover here, I think. Although we shouldn't panic, I'm not sure if we're actually using `close` anywhere (best addressed separately).

libwebauthn/src/transport/nfc/device.rs
```diff
@@ -0,0 +1,139 @@
...
+        Ok(protocols.fido2 || protocols.u2f)
+    }
+
+    inner::<Ctx>(device).await.is_ok()
+}
+
+#[instrument]
+pub async fn list_devices() -> Result<Vec<NfcDevice>, Error> {
+    let mut all_devices = Vec::new();
+    // TODO: Either only allow one backend, OR deduplicate found devices here!
```
AlfioEmanueleFresta: Nit: If possible, create a follow-up issue.

libwebauthn/src/transport/nfc/channel.rs
msirringhaus: yeah, HID also only returns `()` and does nothing. I'm doing the same here now.

libwebauthn/examples/u2f_hid.rs
msirringhaus: I created `u2f_nfc.rs` and `webauthn_nfc.rs` and only compile them if NFC is requested as a feature

libwebauthn/src/transport/nfc/device.rs
msirringhaus: Filed #154 and linked it in the comment.

Issue #14, opened 2022-02-27
Refactoring: Use let-else for errors


AlfioEmanueleFresta:
Currently a lot of errors are handled in a manner similar to:

```rust
let dev = device.open_device(&hidapi)
                .or(Err(Error::Transport(TransportError::ConnectionFailed)))?;
```

These errors are tricky to track down. Now that we're using the `tracing` crate fully, it would be best to rewrite these using `let-else` syntax, ie.:

```
let Ok(dev) = device.open_device(&hidapi) else {
    error!(%err, "Failed to open device");
    return Err(Error::Transport(TransportError::ConnectionFailed));
}
```

If the error is relevant to the message, first consider if the method being called should be instrumented to automatically log errors in their context:
```rust
use tracing::instrument;

#[instrument(skipall, err)]
pub fn open_device(&self, dev: &HidDevice) {
    ...
```

If this is not possible, we can instead use match statements at the point of invocation:

```rust
let dev = match device.open_device(&hidapi) {
    Ok(dev) => dev,
    Err(err) => {
        error!(%err, "Failed to open device");
        return Err(Error::Transport(TransportError::ConnectionFailed));
    }
};
```

PR #153, opened 2025-10-22
Log when CTAP2 bio_enrollment responses return None fields

jkaiwar:
This PR adds `info!` logs in the error paths for a couple of `BioEnrollment` implementation methods, which previously returned `Error::Ctap(CtapError::Other)` silently under some CTAP requests to channels, where the response had `None` fields that were expected to be `Some`.

Maintainers: Are these logs desirable in the context of #14 ? How appropriate are the messages and log-levels I have chosen?

### DIFF
--- a/libwebauthn/src/management/bio_enrollment.rs
+++ b/libwebauthn/src/management/bio_enrollment.rs
@@ -20,7 +20,7 @@ use crate::{
 use async_trait::async_trait;
 use serde_bytes::ByteBuf;
 use std::time::Duration;
-use tracing::info;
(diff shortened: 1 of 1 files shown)


AlfioEmanueleFresta:
I would reserve `error!` to something the library genuinely can't handle, or high-level operation failures.

jkaiwar:
Sounds good @AlfioEmanueleFresta @msirringhaus. I have changed the levels to warn.

AlfioEmanueleFresta (approved):
Thank you @jkaiwar!

PR #155, opened 2025-12-19
Web IDL support 3/N: response JSON serialization

AlfioEmanueleFresta:
⚠️ **AI-Assisted PR**: This pull request was developed with the assistance of an AI coding agent (Claude Opus 4.5).

---

## Summary

This PR is the second in a series (follows #138) and adds JSON serialization for WebAuthn responses - the inverse of the existing JSON request parsing functionality.

### New Response Serialization Infrastructure

- **Add `WebAuthnIDLResponse` trait** - Provides `to_json()` and `to_inner_model()` methods for converting responses to JSON format
- **Add `JsonFormat` enum** - Supports both minified and prettified JSON output
- **Add `ResponseSerializationError`** - Error type for serialization failures

### New JSON Response Models (per WebAuthn Level 3 spec)

- `RegistrationResponseJSON` - For `MakeCredentialResponse` serialization
- `AuthenticationResponseJSON` - For `Assertion` serialization  
- `AuthenticatorAttestationResponseJSON` - Attestation response details
- `AuthenticatorAssertionResponseJSON` - Assertion response details
- `AuthenticationExtensionsClientOutputsJSON` - Extension output serialization

### Request Structure Refactoring

Refactored `MakeCredentialRequest` and `GetAssertionRequest` to enable client data generation on-the-fly:

- **Replaced `hash: Vec<u8>` with `challenge: Vec<u8>`** - Store raw challenge instead of pre-computed hash
- **Added `origin: String` field** - Store origin explicitly
- **Added `cross_origin: Option<bool>` field** - Optional cross-origin flag
- **Removed `client_data: ClientData` field** - No longer needed as separate field
- **Added helper methods**:
  - `client_data()` - Builds `ClientData` internally
  - `client_data_hash()` - Computes SHA-256 hash on demand
  - `client_data_json()` - Returns JSON bytes for response serialization

### Implementation Details

- Implemented `WebAuthnIDLResponse` for `MakeCredentialResponse` with full attestation object CBOR serialization
- Implemented `WebAuthnIDLResponse` for `Assertion` with signature and authenticator data handling
- Support for extension outputs: `credProps`, `hmacCreateSecret`, `hmacGetSecret`, `largeBlob`, `prf`
- Added `Serialize` derives to attestation statement types for CBOR encoding

### Updated Files

- All example files updated to use new request structure
- All test files updated
- CTAP2 model conversions updated to use `client_data_hash()`
- U2F fallback code updated
- WebAuthn preflight logic updated

## Testing

- All 85 existing tests pass
- New unit tests added for response serialization
- Updated `webauthn_json_hid.rs` example to demonstrate JSON response output

## Example Usage

```rust
use libwebauthn::ops::webauthn::{WebAuthnIDLResponse, JsonFormat};

// After a MakeCredential operation
let response = channel.webauthn_make_credential(&request).await?;
let json = response.to_json(&request, JsonFormat::Prettified)?;
println!("{}", json);

// After a GetAssertion operation  
for assertion in &response.assertions {
    let json = assertion.to_json(&request, JsonFormat::Minified)?;
    println!("{}", json);
}
```

### DIFF
--- a/libwebauthn/src/ops/webauthn/client_data.rs
+++ b/libwebauthn/src/ops/webauthn/client_data.rs
@@ -39,3 +39,70 @@ impl ClientData {
         hasher.finalize().to_vec()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn make_client_data(cross_origin: Option<bool>) -> ClientData {
+        ClientData {
+            operation: Operation::GetAssertion,
+            challenge: b"test-challenge".to_vec(),
+            origin: "https://example.org".to_string(),
+            cross_origin,
+            top_origin: None,
+        }
+    }
+
+    #[test]
+    fn test_cross_origin_none_produces_false() {
+        let client_data = make_client_data(None);
+        let json = client_data.to_json();
+        assert!(
+            json.contains("\"crossOrigin\":false"),
+            "Expected crossOrigin:false, got: {}",
+            json
+        );
+    }
+
+    #[test]
+    fn test_cross_origin_false_produces_false() {
+        let client_data = make_client_data(Some(false));
+        let json = client_data.to_json();
+        assert!(
+            json.contains("\"crossOrigin\":false"),
+            "Expected crossOrigin:false, got: {}",
+            json
+        );
+    }
+
+    #[test]
+    fn test_cross_origin_true_produces_true() {
+        let client_data = make_client_data(Some(true));
+        let json = client_data.to_json();
+        assert!(
+            json.contains("\"crossOrigin\":true"),
+            "Expected crossOrigin:true, got: {}",
+            json
+        );
+    }
+
+    #[test]
+    fn test_to_json_format() {
+        let client_data = ClientData {
+            operation: Operation::MakeCredential,
+            challenge: b"DEADCODE".to_vec(),
+            origin: "https://example.org".to_string(),
+            cross_origin: Some(true),
+            top_origin: None,
+        };
+        let json = client_data.to_json();
+
+        // Verify the JSON contains expected structure
+        assert!(json.contains("\"type\":\"webauthn.create\""));
+        assert!(json.contains("\"origin\":\"https://example.org\""));
+        assert!(json.contains("\"crossOrigin\":true"));
+        // Challenge should be base64url encoded
+        assert!(json.contains("\"challenge\":\"REVBRENPREU\"")); // base64url of "DEADCODE"
+    }
+}


libwebauthn/src/ops/webauthn/client_data.rs
```diff
@@ -13,7 +13,8 @@ pub struct ClientData {
 }
 
 impl ClientData {
-    pub fn hash(&self) -> Vec<u8> {
+    /// Returns the canonical JSON representation of the client data.
+    pub fn to_json_bytes(&self) -> Vec<u8> {
```
msirringhaus: I wonder if it would be more versatile to make this `to_json()` and return a String. Those who need the bytes can then call `into_bytes()` on the result.

libwebauthn/examples/webauthn_prf_hid.rs
```diff
@@ -465,7 +468,9 @@ async fn run_failed_test(
 ) {
     let get_assertion = GetAssertionRequest {
         relying_party_id: "example.org".to_owned(),
-        hash: Vec::from(challenge),
+        challenge: Vec::from(challenge),
+        origin: "example.org".to_string(),
+        cross_origin: None,
```
msirringhaus, line 473: Do we have any tests where `cross_origin` is not `None`?

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -39,13 +44,34 @@ pub struct PRFValue {
...
+            origin: self.origin.clone(),
+            cross_origin: self.cross_origin,
+        }
+    }
+
+    pub fn client_data_hash(&self) -> Vec<u8> {
+        self.client_data().hash()
+    }
+
+    pub fn client_data_json(&self) -> Vec<u8> {
```
msirringhaus: Same question as above. More so here: The naming `client_data_json()` would lead me to expect a `String`.

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -362,6 +384,102 @@ pub struct Assertion {
...
 
+impl WebAuthnIDLResponse for Assertion {
+    type InnerModel = AuthenticationResponseJSON;
+    type Context = GetAssertionRequest;
+
+    fn to_inner_model(
+        &self,
+        request: &Self::Context,
+    ) -> Result<Self::InnerModel, ResponseSerializationError> {
+        // Get credential ID - either from credential_id field or from authenticator_data
```
msirringhaus, line 411: I think we do this dance for credential ID in other places, too. Might be worth its own function?
Also, I think there is a special case where we have to use the credential ID from the request. preflight.rs does this IIRC.

libwebauthn/src/ops/webauthn/get_assertion.rs
```diff
@@ -587,4 +701,130 @@ mod tests {
...
+        let assertion = create_test_assertion();
+        let request = create_test_request();
+        let json = assertion.to_json(&request, JsonFormat::default());
+        assert!(json.is_ok());
+
+        let json_str = json.unwrap();
+        let parsed: serde_json::Value = serde_json::from_str(&json_str).unwrap();
+
+        // Verify required fields
+        assert!(parsed.get("id").is_some());
```
msirringhaus: Do we want to also check the contents of the fields, not just if they are `Some()`?

libwebauthn/src/ops/webauthn/idl/response.rs
```diff
@@ -0,0 +1,259 @@
...
+//!
+//! See: https://www.w3.org/TR/webauthn-3/#sctn-public-key-credential-json
+
+use serde::Serialize;
+
+use super::Base64UrlString;
+
+/// JSON output format options.
+#[derive(Debug, Clone, Copy, Default)]
+pub enum JsonFormat {
```
msirringhaus, line 15: Is it intentional to 'reimplement' this, instead of exposing a serde_json-Value, which the user can then use and format according to whatever serde_json offers?
I don't have a strong preference for either way, just wanting to understand the reasoning better, if there is any.

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -40,6 +45,147 @@ pub struct MakeCredentialResponse {
...
+impl WebAuthnIDLResponse for MakeCredentialResponse {
+    type InnerModel = RegistrationResponseJSON;
+    type Context = MakeCredentialRequest;
+
+    fn to_inner_model(
+        &self,
+        request: &Self::Context,
+    ) -> Result<Self::InnerModel, ResponseSerializationError> {
+        // Get credential ID from attested credential data
+        let credential_id_bytes = self
```
msirringhaus, line 68: Same comment as for the `GetAssertion`-case regarding crential IDs.

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -40,6 +45,147 @@ pub struct MakeCredentialResponse {
...
+                client_data_json: Base64UrlString::from(request.client_data_json()),
+                authenticator_data: Base64UrlString::from(authenticator_data_bytes),
+                transports,
+                public_key,
+                public_key_algorithm,
+                attestation_object: Base64UrlString::from(attestation_object_bytes),
+            },
+            authenticator_attachment: None,
+            client_extension_results,
+            r#type: "public-key".to_string(),
```
msirringhaus, line 126: I'm ok with this for now, as we do it in lots of places, but somehow my engineering mind would feel more at ease, if we could make this a transparent enum instead of a hardcoded string in a bunch of places, to avoid typos. But this doesn't have to be in this PR.

libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -631,4 +792,183 @@ mod tests {
...
+        let response = create_test_response();
+        let request = create_test_request();
+        let json = response.to_json(&request, JsonFormat::default());
+        assert!(json.is_ok());
+
+        let json_str = json.unwrap();
+        let parsed: serde_json::Value = serde_json::from_str(&json_str).unwrap();
+
+        // Verify required fields
+        assert!(parsed.get("id").is_some());
```
msirringhaus: Same as with GetAssertion: Can we check the contents as well for those that are easy/obvious?

libwebauthn/src/ops/webauthn/get_assertion.rs
msirringhaus: ah, I just saw https://github.com/linux-credentials/libwebauthn/issues/161 regarding the special case

libwebauthn/src/ops/webauthn/get_assertion.rs
iinuwa: I agree; we should also add the comment that it produces canonical JSON so that users don't have to guess whether they can use the bytes directly or not.

libwebauthn/src/ops/webauthn/make_credential.rs
AlfioEmanueleFresta: I believe the JSON-suffix models should only enforce JSON types (in this case, string). Then, accepting known values is part of the validation that occurs later.

IMO, this is the best way to allow for future credential types: this is a valid JSON model, but it may fail to convert to a strongly-typed enum model.

libwebauthn/src/ops/webauthn/idl/response.rs
AlfioEmanueleFresta: I am of two minds. Returning a `Value`, the caller would need add a dependency on `serde_json` for something as trivial as formatting. Furthermore, as JSON is in a way a technical detail of WebAuthn IDL, I feel like it should be encapsulated within libwebauthn. 

Other than formatting, do we envision any other valid use case where the caller should be manipulating the JSON response?

libwebauthn/examples/webauthn_prf_hid.rs
AlfioEmanueleFresta: Added!

libwebauthn/src/ops/webauthn/client_data.rs
AlfioEmanueleFresta: Fair enough! Addressed.

libwebauthn/src/ops/webauthn/idl/response.rs
msirringhaus: `serde_json` is in the dependency tree anyways, since we pull it it, so adding it shouldn't really hurt users. But I get that it is weird/clunky to return a type that you can only work with, if you add another dependency (or if we re-export some of `serde_json`). 

I'm not sure about manipulating the JSON, but maybe reading? Esp. extensions? But I'm not sure right now.

I'm fine with the way it is right now, I was just curious.

msirringhaus (approved):
lgtm!

libwebauthn/src/ops/webauthn/idl/response.rs
AlfioEmanueleFresta: I agree with you, I'm thinking it would be best to have both methods available, eg. `to_json_value` as well as a convenience method `to_json_string`. Let's do this in a follow-up diff.

PR #165, opened 2026-01-23
Dynamically link hidraw library

iinuwa:
This is up for discussion, but after removing some other dependencies, I noticed that the compilation time of hidapi increased (12.5 out of 56.3 seconds on my machine)

image

hidraw is ubiquitous on the OSes versions we target, so I don't think there's any harm in linking to the system library. (There is a [note in the docs from over 4 years ago](https://github.com/ruabmbua/hidapi-rs/commit/bbb7f3eb4ef91f71cbe2b39b4960019aa110112e#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R50-R51) that "hidraw might be buggy in older kernel versions", but I don't think that applies today.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,10 @@ $ cargo run --example webauthn_cable
 $ cargo run --example u2f_hid
 ```
 
+## Package Requirements
+
+- libhidapi-dev/hidapi-devel
+
 ## Contributing
 
 We welcome contributions!
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Update apt cache
         run: sudo apt-get update
       - name: Install system dependencies
-        run: sudo apt-get install libudev-dev libdbus-1-dev libsodium-dev
+        run: sudo apt-get install libhidapi-dev libudev-dev libdbus-1-dev libsodium-dev
       - name: Build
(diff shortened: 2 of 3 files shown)


kalvdans (approved):
Best would be a rust-only library (hidraw-rs comes up from a search) but change looks fine!

README.md
```diff
@@ -60,6 +60,10 @@ $ cargo run --example webauthn_cable
 $ cargo run --example u2f_hid
 ```
 
+## Package Requirements
+
+- libhidapi-dev/hidapi-devel
```
AlfioEmanueleFresta, line 65: We should add other dependencies here, (libudev, libdbus-1, libsodium), but we can do as a separate PR.

AlfioEmanueleFresta (approved):

README.md
AlfioEmanueleFresta: @copilot+claude-opus-4.6 can you take a look (also fix the build issue if you can see the error and test).

iinuwa:
Hmm. We definitely want to support Ubuntu 24.04, and I'm trying to make this at least compile on Ubuntu 22.04. If this is going to cause issues, then I don't think this is worth much more investigation; we've got higher priority things to work on.

I'm fine to close this.

PR #181, opened 2026-03-05
Make PinNotSet errors an interactive flow

msirringhaus:
Add interactive possibility to 'fix' `PINNotSet` errors, same as `PINRequired`, to help with https://github.com/linux-credentials/credentialsd/pull/135

- For testing, the `webauthn_{hid,nfc}` examples now have a question whether it should use UV preferred or required, to be able to test both (`PINNotSet` obviously only is returned if UV is required)
- Instead of erroring out early as we did before, the new algorithm (hopefully) allows for devices that can handle this internally to continue. Sadly, I don't have such a device to test this.
- Split the `PinManagement`-trait into and internal and public API, which is a bit hacky. 
- Reused `change_pin()` from `PinManagement`, which can both change or set a PIN. Not super pretty, but should do in this case.

This PR should probably stay open until I managed to fully integrated it into the credentialsd PR.

### DIFF
--- a/libwebauthn/src/pin.rs
+++ b/libwebauthn/src/pin.rs
@@ -433,18 +433,136 @@ pub fn hkdf_sha256(salt: Option<&[u8]>, ikm: &[u8], info: &[u8]) -> Result<Vec<u
     Ok(Vec::from(okm))
 }
 
-#[async_trait]
-pub(crate) trait PinManagementInternal {
-    async fn change_pin_internal(
-        &mut self,
-        get_info_response: &Ctap2GetInfoResponse,
-        new_pin: String,
-        timeout: Duration,
-    ) -> Result<(), Error>;
+// Defining a sealed trait for the internal API that is only visible within the crate
+// while not touching/extending the external API
+pub(crate) mod internal {
+    use super::*;
+
+    #[async_trait]
+    pub trait PinManagementInternal {
+        async fn change_pin_internal(
+            &mut self,
+            get_info_response: &Ctap2GetInfoResponse,
+            new_pin: String,
+            timeout: Duration,
+        ) -> Result<(), Error>;
+    }
+
+    #[async_trait]
+    impl<C> PinManagementInternal for C
+    where
+        C: Channel,
+    {
+        async fn change_pin_internal(
+            &mut self,
+            get_info_response: &Ctap2GetInfoResponse,
+            new_pin: String,
+            timeout: Duration,
+        ) -> Result<(), Error> {
+            // If the minPINLength member of the authenticatorGetInfo response is absent, then let platformMinPINLengthInCodePoints be 4.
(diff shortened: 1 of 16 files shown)


libwebauthn/src/webauthn/pin_uv_auth_token.rs
```diff
@@ -539,6 +618,68 @@ mod test {
...
+            status_recv,
+            expected_reasons,
+            pin_answers,
+        ));
+
+        channel.push_command_pair(info_req, info_resp);
+
+        let mut getassertion = create_get_assertion(&info, None);
+
+        // We should early return here right at the start and not send a ClientPIN-request
```
AlfioEmanueleFresta: Nit: This is no longer accurate, right?

libwebauthn/src/webauthn/pin_uv_auth_token.rs
```diff
@@ -576,22 +717,21 @@ mod test {
...
-            // Should be the same as above
-            vec![("clientPin", false)],
+    async fn device_client_pin_not_set_but_uv_required_hanging_up() {
+        let expected_reasons = vec![PinNotSetReason::PinNotSet];
+        let pin_answers = vec![];
+        test_setting_pin(expected_reasons, pin_answers).await;
+    }
+
+    #[tokio::test]
+    async fn device_client_pin_not_set_but_uv_required_try_setting_pin() {
```
AlfioEmanueleFresta, line 727: Should we add a test for the happy path, that verifies the request actually continues?

libwebauthn/src/pin.rs
```diff
@@ -422,7 +434,18 @@ pub fn hkdf_sha256(salt: Option<&[u8]>, ikm: &[u8], info: &[u8]) -> Result<Vec<u
...
+        &mut self,
+        get_info_response: &Ctap2GetInfoResponse,
+        new_pin: String,
+        timeout: Duration,
+    ) -> Result<(), Error>;
+}
+
+#[async_trait]
+#[allow(private_bounds)]
+pub trait PinManagement: PinManagementInternal {
```
AlfioEmanueleFresta, line 566: WDYT about a sealed trait, so we don't need `#[allow(private_bounds)]`?

```rust
mod private {
    pub trait Sealed {}
    impl<C: Channel> Sealed for C {}
}

pub trait PinManagement: private::Sealed {
    async fn change_pin(...) -> Result<(), Error>;
}
```

libwebauthn/src/webauthn/pin_uv_auth_token.rs
msirringhaus: Yes, we probably should. It's rather long, but better to have it

libwebauthn/src/pin.rs
msirringhaus: Yep, probably the more idiomatic approach

AlfioEmanueleFresta (approved):

Issue #187, opened 2026-05-09
Validate rp.id as a registrable suffix of the origin's effective domain


AlfioEmanueleFresta:
## Problem

Currently libwebauthn validates the JSON request's `rp.id` against the origin's effective domain using **strict string equality** (see `from_idl_model` in both `make_credential.rs` and `get_assertion.rs`).

WebAuthn L3 (§5.1.3 step 7, §5.1.7 step 9, §5.1.4) only requires that `rp.id` be a **registrable suffix** of (or equal to) the effective domain. Strict equality rejects valid configurations such as an origin of `https://login.example.com` with `rp.id = "example.com"`.

## Scope

- Replace strict equality in `MakeCredentialRequest::from_idl_model` and `GetAssertionRequest::from_idl_model` with a registrable-suffix check.
- The check needs to be Public Suffix List (PSL) aware so that, for example, `co.uk` is not treated as a registrable domain on its own.
- Applies to both same-origin and cross-origin requests; the comparison is always against the request's effective domain.

## Sequencing

This should land **before** PR #173. The registrable-suffix check is a basic spec correctness fix; related-origin validation is a separate, larger feature on top of it. Doing this one first keeps #173 focused on the §5.11 fetch/validate logic.

## Code-sharing opportunity with #173

PR #173 already pulls in the `publicsuffix` crate and writes domain-label / eTLD helpers for §5.11. Whichever lands first should put those helpers in a shared module so the second PR doesn't duplicate them. If #187 lands first, #173 reuses; if for some reason #173 lands first, #187 extracts.

## Acceptance

- Tests covering: same-domain (passes), parent registrable domain (passes), unrelated registrable domain (fails), eTLD-only (`co.uk`) handled correctly.
- New dep on `publicsuffix` (or whatever PSL crate we settle on), unless #173 already added it.

AlfioEmanueleFresta:
Cc @iinuwa, I believe we already agreed doing this check in libwebauthn - credentialsd checks origin belongs to caller, libwebauthn checks rpid/origin relationship.

iinuwa:
Yup, agreed

PR #188, opened 2026-05-09
refactor(webauthn): use RequestOrigin instead of RelyingPartyId in WebAuthnIDL (fix #185)

### LINKED ISSUE #185: WebAuthnIDL to use Origin rather than RelyingPartyId

AlfioEmanueleFresta:
## Motivation

Closes #185.

`WebAuthnIDL::from_json` currently takes a `&RelyingPartyId`, which forces callers (e.g. credentialsd) to override `request.origin` and `request.cross_origin` after parsing because the bare host is not a valid origin string and we have no place to record the top-level origin. This replaces that parameter with a `RequestOrigin` that carries the actual origin context, so the parsed request comes out correct without a post-parse fixup.

## What changes

- New `Origin` and `RequestOrigin` types in `libwebauthn::ops::webauthn`. `Origin` is a struct with `host: OriginHost` and `port: Option<u16>`; `RequestOrigin` wraps it with an optional `top_origin`. Convenience constructors: `RequestOrigin::new(origin)`, `RequestOrigin::new_cross_origin(origin, top_origin)`, `RequestOrigin::try_from(&str | String)` for one-shot parsing of `"https://host[:port]"`.
- Host validation goes through `url::Host` so we follow the WHATWG URL Standard host parser (domain / IPv4 / bracketed IPv6). Errors are wrapped into a local `HostParseError` / `OriginParseError` so the `url` crate's error type does not leak into the public API.
- `WebAuthnIDL::from_json` and `FromIdlModel::from_idl_model` now take `&RequestOrigin`. The parsed `request.origin` is the full URL string (`"https://example.org"`, no longer the bare host), and `request.top_origin: Option<String>` replaces the old `cross_origin: Option<bool>`.
- `ClientData` drops `cross_origin: Option<bool>` and derives `crossOrigin` in the JSON from `top_origin.is_some()`. `topOrigin` is now emitted when present.
- Bumps libwebauthn to `0.4.0` since the public `from_json` signature and the request struct fields are breaking changes.

## Intentional non-changes

- `Origin` is a plain struct, not an enum. We only support `https`. If we need to support a second scheme later it can become an enum without breaking call-site field access.
- `rp.id` validation against the origin is still strict equality. The spec actually wants a registrable-suffix check (WebAuthn L3 §5.1.3 / §5.1.7); that is tracked separately in #187 and can build on the PSL helpers from #173.

## Test plan

- [x] `cargo build -p libwebauthn` and `cargo build -p libwebauthn --features virt`
- [x] `cargo test --workspace` (149 tests, 14 new origin parser tests)
- [x] `cargo publish --dry-run -p libwebauthn` (no working-tree hacks)

### DIFF
--- a/libwebauthn/src/ops/webauthn/idl/origin.rs
+++ b/libwebauthn/src/ops/webauthn/idl/origin.rs
@@ -1,0 +1,359 @@
+use std::convert::TryFrom;
+use std::fmt::{self, Display};
+use std::str::FromStr;
+
+use url::Host;
+
+#[derive(thiserror::Error, Debug, Clone, PartialEq, Eq)]
+pub enum HostParseError {
+    #[error("empty host")]
+    Empty,
+    #[error("invalid IP address: {0}")]
+    InvalidIp(String),
+    #[error("invalid domain: {0}")]
+    InvalidDomain(String),
+}
+
+#[derive(thiserror::Error, Debug, Clone, PartialEq, Eq)]
+pub enum OriginParseError {
+    #[error("invalid scheme (only https is supported)")]
+    InvalidScheme,
+    #[error("missing host")]
+    MissingHost,
+    #[error("invalid host: {0}")]
+    InvalidHost(#[from] HostParseError),
+    #[error("invalid port: {0}")]
+    InvalidPort(String),
+    #[error("unexpected path or fragment: {0}")]
+    UnexpectedPath(String),
+}
+
+/// Validated host component of an HTTPS origin.
+///
+/// Parsing follows the WHATWG URL Standard host parser via [`url::Host`], which
+/// accepts ASCII / IDNA domains, IPv4 literals, and bracketed IPv6 literals,
+/// and rejects everything else.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OriginHost(String);
+
+impl OriginHost {
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+impl FromStr for OriginHost {
+    type Err = HostParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.is_empty() {
+            return Err(HostParseError::Empty);
+        }
+        match Host::parse(s) {
+            Ok(h) => Ok(OriginHost(h.to_string())),
+            Err(err) => {
+                let msg = err.to_string();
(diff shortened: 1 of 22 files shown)


libwebauthn/src/ops/webauthn/idl/origin.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+
+fn split_host_and_port(s: &str) -> Result<(&str, Option<u16>), OriginParseError> {
+    // IPv6 literals are bracketed. Find the matching `]` first if present so
+    // we don't confuse `:` inside the address with the host/port separator.
+    if let Some(stripped) = s.strip_prefix('[') {
+        let end = stripped.find(']').ok_or_else(|| {
+            OriginParseError::InvalidHost(HostParseError::InvalidIp(s.to_string()))
+        })?;
+        let host = &s[..end + 2]; // include the brackets
```
msirringhaus: Could this panic?

msirringhaus (commented):
I could imagine `http://` being used in some self-hosting env, but I'm ok leaving it out for now. 

I'm in general feeling a bit uneasy about writing all the parsing ourselves, but if there is nothing suitable available (e.g. by Mozilla/servo), then so be it. 

One inline comment needs clarification, otherwise LGTM.

swissbit-jpthiers:
> I could imagine `http://` being used in some self-hosting env, but I'm ok leaving it out for now.

Usage of `http://` for connection to relying party probably breaks the origin binding and thereby fishing-resistance due to missing origin verification capability.

iinuwa:
> I could imagine `http://` being used in some self-hosting env, but I'm ok leaving it out for now.

Yeah, that's a good callout. `http://localhost` (and any port) is considered a secure context, so we should probably also allow it. I'm fine to follow this up in a later PR though.

> Usage of http:// for connection to relying party probably breaks the origin binding and thereby fishing-resistance due to missing origin verification capability.

The RP should still be able to detect the mismatched origin, so phishing resistance would be preservezd. But I agree, this should be restricted to `localhost` for defense in depth.

AlfioEmanueleFresta:
Thank you for your feedback!
* Using `Url::Parse` now, and allowing `http://localhost` and its IPv4/6 variants. This also removes index slicing/panics.
* Also normalising away default ports (https + 443, http + 80).

PR #189, opened 2026-05-09
fix(webauthn): validate rp.id as a registrable domain suffix (fix #187)

### LINKED ISSUE #187: Validate rp.id as a registrable suffix of the origin's effective domain

AlfioEmanueleFresta:
Closes #187. Stacked on #188.

## What

Replaces the strict-equality check between the JSON request's `rp.id` and the origin's effective domain with the spec-correct "registrable domain suffix of or equal to" relation from HTML §6.5 (referenced by WebAuthn L3 §5.1.3 step 7 and §5.1.7 step 9).

Net effect: `rp.id = "example.org"` now works against an origin of `https://login.example.org`, while `rp.id = "co.uk"` against `https://example.co.uk` is still correctly rejected because `co.uk` is a public suffix.

## Implementation

- New helper `is_registrable_domain_suffix_or_equal(rp_id, effective_domain)` in the origin module. Implements the HTML §6.5 algorithm: equality short-circuit, label-aligned suffix check, then PSL-based rejections (rp.id must not be the effective domain's public suffix and must not itself be a public suffix).
- Wired into `from_idl_model` in both `get_assertion.rs` and `make_credential.rs`. The `relying_party_id` stored on the parsed request is the JSON's `rp.id` (when present), which can now legitimately differ from the origin's host.
- Adds the `psl` crate as a dep. PSL data is embedded at compile time, no runtime fetch (unlike the `publicsuffix` v1.5 + `List::fetch()` approach in PR #173). #173 can adopt `psl` later if it wants the same code path.

Related-origin (§5.11) fallback is still a separate concern, tracked by #173 itself. The TODO comment in the code now points there.

## Tests

- 9 unit tests on the helper covering: equality, parent suffix, cousin domains rejected, label-alignment required, eTLD-only rejected (`com` and `co.uk`), longer rp.id rejected, multi-label suffix accepted (`bar.example.com` -> `foo.bar.example.com`), empty rejected.
- 2 new positive integration tests on `from_json` for both `MakeCredentialRequest` and `GetAssertionRequest`: rp.id-as-parent-registrable-suffix accepted, rp.id-as-eTLD rejected.

## Test plan

- [x] `cargo test --workspace` (149 tests)
- [x] `cargo publish --dry-run -p libwebauthn`


AlfioEmanueleFresta:
Closing in favour of a follow-up PR that introduces a PublicSuffixList trait and DatFilePublicSuffixList impl, per the design discussion in #173. The current implementation embeds the PSL via the psl crate, which contradicts the agreed-on design.

PR #190, opened 2026-05-09
fix(webauthn): validate rp.id as a registrable domain suffix (fix #187)

### LINKED ISSUE #187: Validate rp.id as a registrable suffix of the origin's effective domain

AlfioEmanueleFresta:
Closes #187. Stacked on #188.

## What

Replaces the strict-equality check between the JSON request's `rp.id` and the origin's effective domain with the spec-correct "registrable domain suffix of or equal to" relation from HTML §6.5 (referenced by WebAuthn L3 §5.1.3 step 7 and §5.1.7 step 9).

Net effect: `rp.id = "example.org"` now works against an origin of `https://login.example.org`, while `rp.id = "co.uk"` against `https://example.co.uk` is still correctly rejected because `co.uk` is a public suffix.

## PSL strategy

Per the design discussion in #173 (libwebauthn should not bundle and manage its own PSL):

- Adds a `PublicSuffixList` trait with sync `registrable_domain` / `public_suffix` methods.
- Provides a `DatFilePublicSuffixList` impl that reads a Public Suffix List `.dat` file at construction time. `from_system_file()` reads the standard `/usr/share/publicsuffix/public_suffix_list.dat`, kept current by the system package manager.
- No PSL data is bundled in the libwebauthn crate.
- `publicsuffix = "1.5"`. Same crate as #173, just used offline.

## Plumbing into the parsing API

`WebAuthnIDL::from_json` and `FromIdlModel::from_idl_model` now take an additional `psl: &dyn PublicSuffixList` parameter. This is the simplest shape we discussed - explicit, no hidden state on `RequestOrigin`. Yes, it's another breaking signature change on top of #188; the 0.4.0 release is already breaking, so we eat the cost once.

Origin parsing (`Origin::from_str`, `RequestOrigin::try_from`) stays purely syntactic. PSL is only consulted at validation time.

## Commits

1. Add `PublicSuffixList` trait and `DatFilePublicSuffixList` impl.
2. Use `PublicSuffixList` for the registrable-suffix check in `from_idl_model`.

## Tests

- Unit tests on the suffix-check helper using a small `MockPublicSuffixList` (recognises `com`, `co.uk`, `org`, `net`).
- Unit tests on the mock itself.
- 2 new positive integration tests on `from_json` for both `MakeCredentialRequest` and `GetAssertionRequest`: rp.id-as-parent-registrable-suffix accepted, rp.id-as-eTLD rejected.

## Test plan

- [x] `cargo test --workspace` (151 tests)
- [x] `cargo publish --dry-run -p libwebauthn`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1561,6 +1561,17 @@ checksum = "3d3067d79b975e8844ca9eb072e16b31c3c1c36928edf9c6789548c524d0d954"
 
 [[package]]
 name = "idna"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "418a0a6fab821475f634efe3ccc45c013f742efe03d853e8d3355d5cb850ecf8"
+dependencies = [
+ "matches",
+ "unicode-bidi",
+ "unicode-normalization",
+]
+
+[[package]]
+name = "idna"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3b0875f23caa03898994f6ddc501886a45c7d3d62d04d2d90788d47be1b1e4de"
@@ -1791,7 +1802,7 @@ dependencies = [
  "hidapi",
  "hkdf",
  "hmac 0.12.1",
- "idna",
+ "idna 1.1.0",
  "maplit",
  "mockall",
  "nfc1",
@@ -1801,6 +1812,7 @@ dependencies = [
  "num_enum",
  "p256 0.13.2",
  "pcsc",
+ "publicsuffix",
  "qrcode",
  "rand 0.8.6",
  "rustls",
@@ -1941,6 +1953,12 @@ checksum = "d1525a2a28c7f4fa0fc98bb91ae755d1e2d1505079e05539e35bc876b5d65ae9"
 dependencies = [
  "regex-automata",
 ]
+
+[[package]]
+name = "matches"
+version = "0.1.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2532096657941c2fea9c289d370a250971c689d4f143798ff67113ec042024a5"
 
 [[package]]
 name = "memchr"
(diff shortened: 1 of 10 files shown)


msirringhaus (approved):
LGTM overall!
We might want to add a note to the README that the publicsuffix-package is needed on the system during runtime?

AlfioEmanueleFresta:
Thanks @msirringhaus! Updated README.

I also opened #210 because it seems that Fedora comes with the PSL in DAFSA file format (binary) by default, not plaintext. Debian should be fine as is. Adding an extra parser shouldn't be much work.

PR #193, opened 2026-05-09
fix(nfc): require backend feature; rename to nfc-backend-{pcsc,libnfc}

AlfioEmanueleFresta:
Stacked on top of #186.

## Summary

- Fix build failure when only the `nfc` umbrella feature is enabled (no backend).
  - Gate `transport::nfc` on `any(nfc-backend-pcsc, nfc-backend-libnfc)` so the broken code never compiles in the bad combo.
  - Emit a single clear `compile_error!` from `lib.rs` instead of three cascade errors out of `transport/nfc/device.rs`.
- Rename backend features to `nfc-backend-pcsc` and `nfc-backend-libnfc` to make the umbrella relationship explicit.
- Restructure the example list in the top-level README into a transport × protocol table.
  - NFC cells show both backend variants (`nfc-backend-pcsc`, `nfc-backend-libnfc`).
- Extend CI with per-backend coverage so this can't regress.
  - Build `libwebauthn` with only `nfc-backend-pcsc` enabled.
  - Build `libwebauthn` with only `nfc-backend-libnfc` enabled.

## Build matrix verified locally

| Command                                     | Behavior                                |
| ------------------------------------------- | --------------------------------------- |
| `cargo build` (no features)                 | Builds; NFC examples auto-skipped.      |
| `cargo build --features nfc`                | Single clear `compile_error!`.          |
| `cargo build --features nfc-backend-pcsc`   | Builds NFC examples.                    |
| `cargo build --features nfc-backend-libnfc` | Checks pass (link needs `libnfc`).      |

## Test plan

- [x] `cargo build` (default features) succeeds, NFC examples skipped.
- [x] `cargo build --features nfc` produces a single `compile_error!` with the helpful message.
- [x] `cargo build --example webauthn_nfc --features nfc-backend-pcsc` builds.
- [x] On a CI/dev machine with `libnfc` installed, `--features nfc-backend-libnfc` links successfully.
- [x] New CI steps run green on this PR.

PR #194, opened 2026-05-10
refactor(examples): unify, simplify, group examples

AlfioEmanueleFresta:
Stacked on #193. Six small commits that unify and simplify `examples/`.

- Shared helpers in `examples/common/` (`setup_logging`, UV/cable update handlers, `retry_user_errors!` macro, prompt utilities). 16 examples shed their inline copies.
- JSON-first WebAuthn: `webauthn_json_hid` becomes the canonical `webauthn_hid`. `webauthn_nfc` and `webauthn_cable` now consume and emit JSON per the WebAuthn IDL too.
- Naming consistency: `cred_management_hid.rs`, `device_selection_hid.rs`, `prf_replay.rs` (was `prf_test.rs`).
- Folder layout: `examples/{ceremony,features,management}/` with explicit `[[example]]` entries; `cargo run --example <name>` invocations are unchanged.

Net diff: -968 LOC across the example set.

Side effect: fixes a pre-existing off-by-bound in `cred_management` (the credential-selection prompt was bounded against `options.len()` instead of `credlist.len()`).

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo build --examples` with default, `nfc-backend-pcsc`, `nfc-backend-libnfc`
- [ ] Smoke test against HID authenticator
- [ ] Smoke test against NFC reader (PCSC backend)


msirringhaus (approved):
Having to list every example in Cargo.toml might become a bit annoying, but I think it's still better than the previous unstructured approach.
A cleanup like this was indeed long overdue.
Thanks!

PR #196, opened 2026-05-10
fix(webauthn): escape clientDataJSON strings per CCDToString

AlfioEmanueleFresta:
WebAuthn L3 §5.8.1.2 specifies that `clientDataJSON` strings must be escaped via `CCDToString` (a strict subset of ECMA-262 / RFC 8259 string-escape rules). `ClientData::to_json` currently hand-rolls a JSON object via `format!()`, concatenating the `origin` and `topOrigin` strings unescaped, which produces malformed or attacker-injectable JSON when those strings contain `"`, `\`, or U+0000-U+001F.

Empirical repro on the base branch: setting `origin` to `https://example.com","origin":"https://attacker.com` yields a document with two `origin` keys.

Stacked on top of #190 (which is stacked on #188 / #186). The `RequestOrigin` plumbing introduced by #188 already correctly populates `ClientData::origin` and `ClientData::top_origin` from the URL origin; that work is preserved unchanged here.

## Changes

- Replace hand-rolled `format!()` in `ClientData::to_json` with a `#[derive(Serialize)]` wire struct that mirrors the spec field order (`type`, `challenge`, `origin`, `topOrigin?`, `crossOrigin`). `serde_json`'s RFC 8259 string encoder is a strict superset of `CCDToString`.
- `topOrigin` is omitted from the output entirely when absent (via `skip_serializing_if`), preserving the existing behaviour.
- Regression tests for: origin containing `"`, origin containing `\`, origin containing U+0000-U+001F control characters, hostile `topOrigin`, spec field order with and without `topOrigin`, and a full round-trip.

## Test plan

- [x] `cargo build`
- [x] `cargo test --lib` (159 passed, including 11 in `client_data::tests`)
- [x] `cargo fmt --check`
- [x] `cargo clippy --lib --all-features --no-deps -- -D warnings`

Refs: WebAuthn L3 §5.8.1.2 (`CCDToString`); audit dossiers D-01, D-02, D-04, B-10, H-04, H-05.

### DIFF
--- a/libwebauthn/src/ops/webauthn/client_data.rs
+++ b/libwebauthn/src/ops/webauthn/client_data.rs
@@ -1,6 +1,6 @@
 use crate::ops::webauthn::Operation;
 
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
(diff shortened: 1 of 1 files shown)


msirringhaus (approved):
LGTM!

PR #198, opened 2026-05-10
fix(webauthn): largeBlob.read no longer leaks largeBlobKey to RP

AlfioEmanueleFresta:
Per [WebAuthn L3 sec. 10.1.5 (largeBlob extension)](https://www.w3.org/TR/webauthn-3/#sctn-large-blob-extension), the relying party expects the `blob` output to be the decrypted plaintext blob payload, fetched by the platform via the CTAP [`authenticatorLargeBlobs`](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#authenticatorLargeBlobs) command using the per-credential `largeBlobKey` as an AES-256-GCM key. The library was instead writing the raw `largeBlobKey` into `blob` and never calling `authenticatorLargeBlobs`. That means an RP receives the AES key itself (not the blob), and if the RP can also read the device's `largeBlobArray` (publicly readable region of the authenticator over CTAP), it can decrypt and forge entries.

Until `authenticatorLargeBlobs` is wired up (follow-up PR #206), the safe behaviour is to set `large_blob.blob = None`. The CTAP-level model keeps the field so the follow-up can use it.

## Changes

- Stop routing `largeBlobKey` into the WebAuthn `large_blob.blob` output.
- The CTAP-level `Ctap2GetAssertionResponse.large_blob_key` remains so the next PR can use it.
- Regression test asserts the WebAuthn response no longer contains the key.

## References

- [WebAuthn L3 sec. 10.1.5: Large blob storage extension (`largeBlob`)](https://www.w3.org/TR/webauthn-3/#sctn-large-blob-extension)
- [CTAP 2.1 sec. 6.10: `authenticatorLargeBlobs` (0x0C)](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#authenticatorLargeBlobs)


msirringhaus (approved):
Yep, that's right. I was actually planning on finalizing my LargeBlob-PR, that would have fixed that, too, but that got delayed by other stuff. This is a good workaround until then

PR #199, opened 2026-05-10
fix(ctap2): check status code in getNextAssertion

AlfioEmanueleFresta:
Every other CTAP2 method already checks the status byte and surfaces non-OK as a typed `CtapError`; `ctap2_get_next_assertion` was the lone exception. Without the check, a device returning `CTAP2_ERR_NOT_ALLOWED` (the spec-defined error when no further assertion is available within the 30-second window) is silently masked as `InvalidDeviceResponse`; worse, any bytes the device puts after a non-OK status byte (per CTAP 2.2 §8 they are undefined) would be parsed as a valid assertion.

## Changes
- Add the standard `match cbor_response.status_code` block at the top of `ctap2_get_next_assertion` so non-OK status codes surface as `Error::Ctap(<variant>)` instead of being silently parsed.
- Add regression tests covering status-code propagation across the broader set of CTAP2 methods (`ctap2_get_info`, `ctap2_make_credential`, `ctap2_get_assertion`, `ctap2_get_next_assertion`, `ctap2_client_pin`, `ctap2_selection`, `ctap2_authenticator_config`, `ctap2_bio_enrollment`, `ctap2_credential_management`) so a future contributor can't accidentally regress the pattern.

## References
- CTAP 2.2 §8 Message Encoding: https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#message-encoding
- CTAP 2.2 §6.3 authenticatorGetNextAssertion: https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#authenticatorGetNextAssertion

### DIFF
--- a/libwebauthn/src/proto/ctap2/protocol.rs
+++ b/libwebauthn/src/proto/ctap2/protocol.rs
@@ -282,7 +282,14 @@ where
 mod tests {
     use std::time::Duration;
 
+    use serde_bytes::ByteBuf;
+
     use crate::proto::ctap2::cbor::{CborRequest, CborResponse};
+    use crate::proto::ctap2::model::{
+        Ctap2AuthenticatorConfigRequest, Ctap2BioEnrollmentRequest, Ctap2ClientPinRequest,
+        Ctap2CredentialManagementRequest, Ctap2GetAssertionRequest, Ctap2MakeCredentialRequest,
+        Ctap2PinUvAuthProtocol,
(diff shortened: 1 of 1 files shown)


msirringhaus (approved):
LGTM!

PR #200, opened 2026-05-10
fix(ctap1): emit Le on U2F APDUs to support case 4

AlfioEmanueleFresta:
U2F REGISTER and AUTHENTICATE produce a response payload, so per FIDO U2F Raw Message Formats v1.2 section 3.1 the request APDU must include `Le` (omitting `Le` is reserved for instructions not expected to yield response bytes). libwebauthn was setting `response_max_length` on the `ApduRequest` but both encoders silently dropped it:

- `ApduRequest::raw_long` stopped after the `Lc + data` field. With `response_max_length = Some(0x100)` the byte stream came out as a Case 3 extended APDU (no `Le`).
- The NFC `From<&ApduRequest> for Command` used `Command::new_with_payload`, which does not carry `Le` at all.

Strict authenticators reject Case 3 register/authenticate. The in-tree virtual `fido-authenticator` is lenient about it, so CI never caught the gap.

## Changes
- `ApduRequest::raw_long` now appends a 2-byte big-endian `Le` when `response_max_length` is `Some(_)`, matching the extended-length encoding `Le1 Le2` from FIDO U2F Raw Message Formats v1.2 section 3.1.3. Values `>= 65536` collapse to `0x0000` (the spec's wildcard for `Ne = 65536`). This is the path used by HID, BLE, and caBLE, all of which mandate extended-length encoding (FIDO U2F HID v1.2 section 2.4; FIDO U2F BT v1.2 sections 5.1 and 5.2).
- NFC `From<&ApduRequest> for Command` switches to `Command::new_with_payload_le`, so `response_max_length` is propagated through `apdu-core`. With `longer_payloads` disabled (the default in this crate), `apdu-core` writes `Le` as a single byte; for the typical request with `response_max_length = 256` this yields `Le = 0x00`, which is short-form Case 4 and valid under FIDO U2F NFC v1.2 section 3 ("either short or extended length APDU encoding is allowed").
- Unit tests cover the Case 3 / Case 4 encoding for both `raw_long` and the NFC encoder, including the wildcard `Le = 0x0000`.

## Spec references
- FIDO U2F Raw Message Formats v1.2, sections 3 (framing), 4 (REGISTER), 5 (AUTHENTICATE): https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-raw-message-formats-v1.2-ps-20170411.html
- FIDO U2F HID Protocol v1.2, section 2.4: https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-hid-protocol-v1.2-ps-20170411.html
- FIDO U2F NFC Protocol v1.2, section 3: https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-nfc-protocol-v1.2-ps-20170411.html
- FIDO U2F BLE Protocol v1.2, sections 5.1, 5.2: https://fidoalliance.org/specs/fido-u2f-v1.2-ps-20170411/fido-u2f-bt-protocol-v1.2-ps-20170411.html

### DIFF
--- a/libwebauthn/src/proto/ctap1/apdu/request.rs
+++ b/libwebauthn/src/proto/ctap1/apdu/request.rs
@@ -106,6 +106,14 @@ impl ApduRequest {
             raw.write_u24::<BigEndian>(0)?;
         }
 
+        // Per ISO 7816-4 and FIDO U2F Raw Message Formats §3, §4: when a
+        // response is expected, append a 2-byte extended-length Le.
+        // Le=0x0000 is the wildcard meaning "up to 65536 bytes".
+        if let Some(le) = self.response_max_length {
+            let le_field = if le >= 0x1_0000 { 0u16 } else { le as u16 };
+            raw.write_u16::<BigEndian>(le_field)?;
+        }
+
         Ok(raw)
     }
 }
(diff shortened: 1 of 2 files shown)


msirringhaus (approved):
Looks good. Has it been tested with hardware?

PR #201, opened 2026-05-10
fix(webauthn): refuse U2F downgrade when credProtect enforce or largeBlob required

AlfioEmanueleFresta:
When the platform negotiates CTAP1/U2F (because the authenticator does not speak CTAP2), `is_downgradable()` was permitting the downgrade even for requests that asked for capabilities U2F cannot provide: an RP that set `enforceCredentialProtectionPolicy=true` with `userVerificationRequired` would end up with an unprotected U2F credential, and an RP that asked for `largeBlob: required` would get a credential with no blob storage. Both are silent capability strips with no signal to the RP.

This change makes `is_downgradable()` refuse the downgrade for those two cases, and surfaces the same enforcement for `largeBlob: required` on the FIDO2 path when the authenticator does not advertise the extension (previously a `warn!`-only path), matching the existing `credProtect` handling immediately above it.

## Changes
- `MakeCredentialRequest::is_downgradable()` returns false when `extensions.cred_protect.enforce_policy` is set with a policy stricter than `UserVerificationOptional`.
- `MakeCredentialRequest::is_downgradable()` returns false when `extensions.large_blob.support == Required`.
- `Ctap2MakeCredentialsRequestExtensions::from_webauthn_request` returns `Err(CtapError::UnsupportedExtension)` for `largeBlob: required` against devices without the `largeBlobs` option.

## References
- WebAuthn Level 3, [10.1.5 Large blob storage extension (largeBlob)](https://www.w3.org/TR/webauthn-3/#sctn-large-blob-extension)
- CTAP 2.1, [10.2 Using the CTAP2 authenticatorMakeCredential Command with CTAP1/U2F authenticators](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#u2f-authenticatorMakeCredential-interoperability)
- CTAP 2.1, [12.1 Credential Protection (credProtect)](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-credProtect-extension)

### DIFF
--- a/libwebauthn/src/ops/webauthn/make_credential.rs
+++ b/libwebauthn/src/ops/webauthn/make_credential.rs
@@ -599,6 +599,32 @@ impl DowngradableRequest<RegisterRequest> for MakeCredentialRequest {
             return false;
         }
 
+        // Enforced credProtect with a non-default policy cannot be honoured by U2F.
+        if let Some(cred_protect) = self
+            .extensions
+            .as_ref()
+            .and_then(|e| e.cred_protect.as_ref())
+        {
+            if cred_protect.enforce_policy
+                && cred_protect.policy != CredentialProtectionPolicy::UserVerificationOptional
+            {
+                debug!("Not downgradable: request enforces a non-default credProtect policy");
+                return false;
+            }
+        }
+
+        // U2F has no large-blob storage.
+        if matches!(
+            self.extensions
+                .as_ref()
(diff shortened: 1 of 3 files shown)


msirringhaus (approved):
LGTM

PR #202, opened 2026-05-10
fix(ble): notifications, write-with-response, LE secure connections check

AlfioEmanueleFresta:
Three FIDO-BLE protocol-level fixes, one per commit, grounded in CTAP 2.2 [§11.4](https://fidoalliance.org/specs/fido-v2.2-ps-20250714/fido-client-to-authenticator-protocol-v2.2-ps-20250714.html#ble):

- **Consume `fidoStatus` notifications instead of GATT Read.** `fidoStatus` is notify-only; spec-conformant peripherals reject Read with `Not Permitted`. `Connection::new` now subscribes and `frame_recv` awaits the next notification. On timeout it emits a best-effort `BleCommand::Cancel` and returns `TransportError::Timeout`.
- **Use Write Request, not Write Command, for spec-declared Write characteristics.** `Write-Without-Response` masked ATT errors (e.g. encryption-required) and was rejected outright by some authenticators. A new `write_type_for()` helper picks `WithoutResponse` only when that is the sole property advertised.
- **Refuse unbonded LE links.** §11.4.3 requires the link to be encrypted before any FIDO message is exchanged. The library queries bluez's `org.bluez.Device1.{Paired,Bonded}` via DBus on `spawn_blocking`; non-bonded devices are refused with `ConnectionFailed` before any BLE connect. On non-bluez backends or when DBus is unreachable, the check falls through.

A follow-up commit fixes two pre-existing BLE enumeration bugs that only show up for peripherals not already actively connected in the current btleplug session, and adds a `webauthn_ble` example mirroring `webauthn_hid`.

## Test plan

- [x] `cargo test --lib` (174 tests), `cargo fmt --check`, `cargo clippy --lib --all-features -- -D warnings`.
- [x] **Bonded, FIDO2**: paired a Thetis FIDO2 BLE key with `bluetoothctl pair`, ran `cargo run --example webauthn_ble`; MakeCredential + GetAssertion succeeded end-to-end.
- [x] **Unbonded**: stripped the bond, ran the same example against the Thetis; library refused with a clear WARN log and no BLE link was opened.


msirringhaus (approved):
With my limited knowledge around BLE: LGTM

PR #203, opened 2026-05-10
fix(security): bound panics on hostile authenticator inputs

AlfioEmanueleFresta:
A misbehaving or hostile authenticator can crash the platform process through five reachable code paths, all of which trust device-supplied lengths and indices that the spec does not actually guarantee. The crate-wide `deny(clippy::panic)` lint does not catch any of them, since the panics originate from `assert!` macros (which `clippy::panic` does not see) or from `.into()` and slice-indexing conversions inside transitive dependencies (`generic-array`, etc.).

## Changes

- **PIN/UV ECDH.** The peer's COSE EC2 key x/y are accepted by `cosey` at any length 0..=32, but the subsequent `EncodedPoint::from_affine_coordinates` requires exactly 32 each. Without this fix, a malformed `getKeyAgreement` response (e.g. a 31-byte x coordinate) panics the very first ECDH step of every PIN/UV ceremony. The fix validates length up front.
- **PIN/UV protocol 2 slicing.** `&key[..32]` / `&key[32..]` panic when the key is shorter than 32 bytes. The relevant entry point is `authenticate(uv_auth_token, clientDataHash)` immediately after the platform decrypts the device-supplied `pinUvAuthToken`; an authenticator returning an IV-only ciphertext yields a zero-byte plaintext, which used to panic at the slicing step. The fix uses checked slicing and validates the token length (16 for PUAP1, 32 for PUAP2) at the decryption boundary.
- **`uv_operation` assertion.** `assert!(self.option_enabled("clientPin"))` fired when an authenticator advertised `pinUvAuthToken: true` with no `clientPin` and no `uv` option. The new code returns `OnlyForSharedSecret` for that case rather than panicking.
- **caBLE `decrypt_frame`.** The final byte of a Noise-decrypted frame is the padding length; on a zero-length plaintext the subtraction underflows and panics. The fix returns `InvalidFraming` on empty or under-padded frames.
- **U2F register response.** A production `assert!(cose_encoded_public_key.len() == 77)` panicked if `cosey` produced a non-canonical length (a real risk on future `cosey` versions or unusual EC keys). Replaced with a typed error return.

Each fix has a regression test in the same module that constructs the malformed input and asserts the error variant.

## Spec references

- CTAP 2.2 §6.5 authenticatorClientPIN: https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#authenticatorClientPIN
- CTAP 2.2 §11.5 Hybrid transports (caBLE): https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#hybrid
- WebAuthn Level 3: https://www.w3.org/TR/webauthn-3/

### DIFF
--- a/libwebauthn/src/transport/cable/tunnel.rs
+++ b/libwebauthn/src/transport/cable/tunnel.rs
@@ -523,6 +523,31 @@ async fn connection_recv_binary_frame(message: Message) -> Result<Option<Vec<u8>
     }
 }
 
+/// Strip the trailing padding-length byte and `padding_len` bytes of padding
+/// from a decrypted Noise transport frame, returning `InvalidFraming` on an
+/// empty plaintext or a declared padding length that exceeds the frame.
+fn strip_frame_padding(mut decrypted_frame: Vec<u8>) -> Result<Vec<u8>, Error> {
+    let padding_len = match decrypted_frame.last() {
+        Some(&b) => b as usize,
+        None => {
+            error!("Decrypted frame is empty; cannot read padding length");
+            return Err(Error::Transport(TransportError::InvalidFraming));
+        }
+    };
+    let new_len = decrypted_frame
+        .len()
+        .checked_sub(padding_len + 1)
+        .ok_or_else(|| {
+            error!(
+                frame_len = decrypted_frame.len(),
+                padding_len, "Padding length exceeds frame length"
+            );
+            Error::Transport(TransportError::InvalidFraming)
+        })?;
+    decrypted_frame.truncate(new_len);
+    Ok(decrypted_frame)
+}
+
 async fn decrypt_frame(
     encrypted_frame: Vec<u8>,
     noise_state: &mut TunnelNoiseState,
@@ -543,8 +568,7 @@ async fn decrypt_frame(
         }
     }
 
-    let padding_len = decrypted_frame[decrypted_frame.len() - 1] as usize;
-    decrypted_frame.truncate(decrypted_frame.len() - (padding_len + 1));
+    let decrypted_frame = strip_frame_padding(decrypted_frame)?;
     trace!(
         ?decrypted_frame,
         decrypted_frame_len = decrypted_frame.len(),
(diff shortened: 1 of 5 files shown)


libwebauthn/src/transport/cable/tunnel.rs
```diff
@@ -523,6 +523,31 @@ async fn connection_recv_binary_frame(message: Message) -> Result<Option<Vec<u8>
     }
 }
 
+/// Strip the trailing padding-length byte and `padding_len` bytes of padding
```
msirringhaus, line 526: I had to read this sentence like 4 times 😂 
(No need to change it, though. It was just funny to me.)

libwebauthn/src/webauthn/pin_uv_auth_token.rs
```diff
@@ -359,6 +359,22 @@ where
                 let uv_auth_token =
                     uv_proto.decrypt(&shared_secret, &encrypted_pin_uv_auth_token)?;
 
+                // pinUvAuthToken is 16 bytes for PUAP1 and 32 bytes for PUAP2.
+                // Reject a shorter token before it is used as a key downstream.
+                let min_token_len = match uv_proto.version() {
+                    Ctap2PinUvAuthProtocol::One => 16,
```
msirringhaus: I'm not entirely sure this logic is correct.

Under section PUAP1, the spec says:
> [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#puatoken-pinuvauthtoken), a random, opaque byte string that MUST be either 16 or 32 bytes long.

while under PUAP2, it says:
> The length of the [pinUvAuthToken](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#puatoken-pinuvauthtoken) for PIN/UV auth protocol two MUST be 32 bytes.

This code would allow under PUAP1 a token of 20 bytes length. I think it has to check for 32 bytes equality always and optionally for 16 bytes, if we are doing PUAP1.

libwebauthn/src/webauthn/pin_uv_auth_token.rs
```diff
@@ -838,7 +854,7 @@ mod test {
         .unwrap();
         // We do here what the device would need to do, i.e. generate a new random
         // pinUvAuthToken (here all 5's), then encrypt it using the shared_secret.
-        let token = [5; 32];
+        let token = [5; 16];
```
msirringhaus, line 875: While puap1 would allow for 32 as well, it is probably not a bad idea to go for 16 here, just to test that both lengths works.

msirringhaus (changes requested):
I think there is a bug in the token length checks

libwebauthn/src/webauthn/pin_uv_auth_token.rs
AlfioEmanueleFresta: Agreed, good catch. Fixed in 605f30e, also added a test.

PR #211, opened 2026-05-12
fix(webauthn): allow PRF inputs of any length (fix #209)

### LINKED ISSUE #209: PRF Inputs incorrectly required to be 32 bytes
As far as I can tell, WebAuthn PRF salt inputs do not have any required length in WebAuthn, only salt outputs. (CTAP2 `hmac-secret` salt inputs are required to be 32 bytes, but there is not an explicit requirement in the WebAuthn spec.)

We use the same type for PRF input and output values `PRFValue`, which requires them to be 32 bytes. We should split these into `PrfInputValue` and `PrfOutputValue` where the former allows any length (greater than 1?).

AlfioEmanueleFresta:
Closes #209.

Split `PRFValue` into `PrfInputValue` (`Vec<u8>`, variable-length salt inputs) and `PrfOutputValue` (`[u8; 32]`, fixed-length hmac-secret outputs).

Spec: [W3C WebAuthn L3 §10.1.4](https://www.w3.org/TR/webauthn-3/#prf-extension) — PRF salt inputs are `BufferSource`s of any length; the client hashes them via `SHA-256("WebAuthn PRF" || 0x00 || input)` before CTAP2 hmac-secret. The hashing was already in place; only the input type was over-constrained.

## Tests
- Unit: variable-length (1/16/31/33/64/256 bytes), empty input, `eval_by_credential` variable-length, plus un-`#[ignore]`'d `test_request_from_json_prf_extension`.
- Virtual device: `test_webauthn_prf_variable_length_input` exercises 0/7/100-byte salts end-to-end and asserts PRF determinism.

### DIFF
--- a/libwebauthn-tests/tests/prf.rs
+++ b/libwebauthn-tests/tests/prf.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 
 use libwebauthn::ops::webauthn::{
     GetAssertionRequest, GetAssertionRequestExtensions, MakeCredentialPrfInput,
-    MakeCredentialPrfOutput, MakeCredentialsRequestExtensions, PRFValue, PrfInput,
+    MakeCredentialPrfOutput, MakeCredentialsRequestExtensions, PrfInput, PrfInputValue,
 };
 use libwebauthn::pin::PinManagement;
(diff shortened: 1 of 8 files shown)


msirringhaus (approved):
LGTM!

PR #213, opened 2026-05-12
feat(webauthn): force UV=required when PRF extension is requested (fix #183)

### LINKED ISSUE #183: Require UV for WebAuthn PRF requests
It seems that the WebAuthn spec will soon require clients to upgrade to `userVerification: "required"` when PRF is requested. (Cf. https://github.com/w3c/webauthn/issues/2337#issuecomment-4390856930.) I think this is a good direction, and we should enforce this for WebAuthn requests with PRF.

(Note that this is for WebAuthn PRF only. For CTAP2-level requests, the `hmac-secret` can still have separate seeds for UV vs. non-UV, and that should not be changed.)

AlfioEmanueleFresta:
Closes #183. Implements the W3C direction in w3c/webauthn#2337: presence of the PRF extension on `create()` or `get()` transparently forces `userVerification = "required"` before the request reaches the authenticator. The CTAP2-level `hmac-secret` extension is unchanged.

### Behaviour changes for callers

- `Discouraged` / `Preferred` + PRF now triggers UV. On no-PIN devices this enters the interactive PIN-setup flow (`PinNotSet`).
- U2F-only devices: PRF requests now error with `NegotiationFailed` instead of silently dropping PRF (U2F has no PRF).

### Notes

- Enforcement is at the public WebAuthn API entry (`webauthn.rs`), so both JSON-IDL and direct-struct callers are covered.
- Unit tests cover the predicate truth table.
- Integration tests cover registration upgrade, assertion upgrade (with a `clear_uv_auth_token_store()` to force fresh UV), and a negative case proving the upgrade does not fire without PRF.

### DIFF
--- a/libwebauthn-tests/tests/prf.rs
+++ b/libwebauthn-tests/tests/prf.rs
@@ -8,7 +8,7 @@ use libwebauthn::ops::webauthn::{
 use libwebauthn::pin::PinManagement;
 use libwebauthn::proto::ctap2::{Ctap2PinUvAuthProtocol, Ctap2PublicKeyCredentialDescriptor};
 use libwebauthn::transport::hid::channel::HidChannel;
-use libwebauthn::transport::{Channel, Device};
+use libwebauthn::transport::{Channel, Ctap2AuthTokenStore, Device};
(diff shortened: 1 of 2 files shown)


iinuwa (approved):
Thanks!

PR #215, opened 2026-05-12
feat(psl): add DAFSA-format Public Suffix List reader

### LINKED ISSUE #210: Add a DAFSA-format Public Suffix List reader
`DatFilePublicSuffixList` (added in #190) reads only the text `.dat` format via the `publicsuffix` crate. On a default Fedora install the `.dat` file is not present, but the binary `.dafsa` file at `/usr/share/publicsuffix/public_suffix_list.dafsa` is. With libwebauthn planned to ship by default on Fedora, the system loader will fail out of the box.

## What ships where

| Distro | `.dat` shipped | `.dafsa` shipped | On default install |
|---|---|---|---|
| Debian/Ubuntu | yes ([file listing](https://packages.debian.org/sid/all/publicsuffix/filelist)) | yes (same package) | yes, `libpsl` Recommends `publicsuffix` |
| Fedora | only via [`publicsuffix-list`](https://packages.fedoraproject.org/pkgs/publicsuffix-list/publicsuffix-list/) | only via [`publicsuffix-list-dafsa`](https://packages.fedoraproject.org/pkgs/publicsuffix-list/publicsuffix-list-dafsa/) | only `-dafsa` (`libpsl` Requires it) |
| Arch | yes ([file listing](https://archlinux.org/packages/extra/any/publicsuffix-list/files/)) | not packaged | no |

On Fedora, `libpsl` (transitively pulled in by curl, wget, git) requires `publicsuffix-list-dafsa` but not `publicsuffix-list`, so nearly every Fedora system has the binary `.dafsa` and almost none have the text `.dat`.

## Proposal

Add a DAFSA reader behind the existing `PublicSuffixList` trait. Open question for implementation time: ship a combined reader that probes `.dafsa` then falls back to `.dat`, or DAFSA-only and drop the `.dat` path. Either way the new code is the DAFSA parser itself.

The libpsl DAFSA format is documented in upstream libpsl (`src/psl-make-dafsa.py`, `src/psl-load-dafsa.c`). A safe-Rust reader is a self-contained ~150-300 LoC task plus tests. The `publicsuffix` crate v1.5 we currently use does not support DAFSA.

## Why not bundle a snapshot

PSL freshness needs to track the system package manager, not crate releases. The point of `DatFilePublicSuffixList` is to delegate that update cadence to the OS. See discussion in #173.

AlfioEmanueleFresta:
Closes #210.

Adds a safe-Rust reader for libpsl's binary `.dafsa` PSL file and an auto-detecting `SystemPublicSuffixList::auto()` loader that probes `.dafsa` then `.dat`. Fixes the Fedora default-install gap where only `.dafsa` is shipped.

The DAFSA lookup is a port of [`LookupStringInFixedSet`](https://github.com/rockdaboot/libpsl/blob/master/src/lookup_string_in_fixed_set.c); format is described in [`psl-make-dafsa`](https://github.com/rockdaboot/libpsl/blob/master/src/psl-make-dafsa). No new runtime deps.

## Test plan

- [x] CI passes (clippy + fmt + workspace tests) with the new \`apt-get install publicsuffix\` step
- [x] Gated integration test \`psl::system::tests::system_psl_loads_and_resolves_common_suffixes\` runs in CI (\`LIBWEBAUTHN_PSL_SYSTEM_TEST=1\`) against Debian's real PSL files
- [x] Verified locally on Fedora that \`SystemPublicSuffixList::auto()\` loads the system \`.dafsa\` and resolves IDN suffixes in punycode form (e.g. \`xn--55qx5d.hk\`)

### DIFF
--- a/libwebauthn/src/ops/webauthn/psl/dafsa.rs
+++ b/libwebauthn/src/ops/webauthn/psl/dafsa.rs
@@ -1,0 +1,405 @@
+//! libpsl binary `.dafsa` format Public Suffix List reader.
+//!
+//! Format reference: <https://github.com/rockdaboot/libpsl/blob/master/src/psl-make-dafsa>
+//! (writer) and <https://github.com/rockdaboot/libpsl/blob/master/src/lookup_string_in_fixed_set.c>
+//! (reader). The on-disk file is a 16-byte ASCII header (`.DAFSA@PSL_<ver>` padded
+//! to 16 bytes with spaces and terminated by LF) followed by a byte-coded DAFSA.
+//! Only version 0 exists today.
+
+use std::path::{Path, PathBuf};
+
+use super::PublicSuffixList;
+
+const MAGIC: &[u8] = b".DAFSA@PSL_";
+const HEADER_LEN: usize = 16;
+
+const FLAG_EXCEPTION: u8 = 1 << 0;
+const FLAG_WILDCARD: u8 = 1 << 1;
+
+#[derive(thiserror::Error, Debug)]
+pub enum DafsaFileLoadError {
+    #[error("io error: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("file too small to contain a valid DAFSA header")]
+    Truncated,
+    #[error("not a libpsl DAFSA file (missing or malformed magic)")]
+    BadMagic,
+    #[error("unsupported DAFSA version: {0}")]
+    UnsupportedVersion(u32),
+}
+
+/// Standard system path for the binary `.dafsa` Public Suffix List shipped
+/// by libpsl's distribution package (e.g. `publicsuffix-list-dafsa` on
+/// Fedora, the `publicsuffix` package on Debian/Ubuntu).
+pub const SYSTEM_PSL_DAFSA_PATH: &str = "/usr/share/publicsuffix/public_suffix_list.dafsa";
+
+/// `PublicSuffixList` implementation backed by libpsl's binary `.dafsa` file.
+pub struct DafsaFilePublicSuffixList {
+    graph: Vec<u8>,
+    source: PathBuf,
+}
+
+impl std::fmt::Debug for DafsaFilePublicSuffixList {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DafsaFilePublicSuffixList")
+            .field("source", &self.source)
+            .field("graph_bytes", &self.graph.len())
+            .finish()
+    }
+}
+
+impl DafsaFilePublicSuffixList {
+    /// Reads a libpsl `.dafsa` file from `path`.
+    pub fn from_path(path: impl AsRef<Path>) -> Result<Self, DafsaFileLoadError> {
+        let path = path.as_ref();
+        let bytes = std::fs::read(path)?;
+        let graph = parse_header(&bytes)?;
+        Ok(Self {
+            graph,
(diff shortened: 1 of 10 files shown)


libwebauthn/src/ops/webauthn/psl/dafsa.rs
```diff
@@ -0,0 +1,413 @@
...
+        .ok_or(DafsaFileLoadError::Truncated)?;
+    if !header.starts_with(MAGIC) {
+        return Err(DafsaFileLoadError::BadMagic);
+    }
+    if header.last() != Some(&b'\n') {
+        return Err(DafsaFileLoadError::BadMagic);
+    }
+    let version_field = header
+        .get(MAGIC.len()..HEADER_LEN - 1)
+        .ok_or(DafsaFileLoadError::BadMagic)?;
```
msirringhaus: Shouldn't this be `DafsaFileLoadError::Truncated`?

libwebauthn/src/ops/webauthn/psl/dafsa.rs
```diff
@@ -0,0 +1,413 @@
...
+    let digit_count = version_field
+        .iter()
+        .take_while(|b| b.is_ascii_digit())
+        .count();
+    let version_digits = version_field
+        .get(..digit_count)
+        .filter(|digits| !digits.is_empty())
+        .ok_or(DafsaFileLoadError::BadMagic)?;
```
msirringhaus: Isn't the header bad and not the magic here (i.e. do we need a `BadHeader`-enum variant as well?)? Same for the usages below

libwebauthn/src/ops/webauthn/psl/system.rs
```diff
@@ -0,0 +1,109 @@
...
+            Inner::Dat(d) => f.debug_tuple("SystemPublicSuffixList").field(d).finish(),
+        }
+    }
+}
+
+impl SystemPublicSuffixList {
+    /// Probes the standard system paths and loads the first format found.
+    ///
+    /// Order: [`SYSTEM_PSL_DAFSA_PATH`] then [`SYSTEM_PSL_PATH`]. The DAFSA
+    /// path is preferred because on Fedora it is the only file that ships
```
msirringhaus, line 48: Why is DAFSA preferred? DAT seems to be far less involved and CPU-intensive to parse? On systems with only DAFSA, we could still prefer DAT with the logic below just inverted.

msirringhaus (approved):
In principle ok. Some smaller nitpicks and questions, but ok as is.

libwebauthn/src/ops/webauthn/psl/system.rs
AlfioEmanueleFresta: Measured on Fedora: DAFSA ~170 KB resident vs DAT ~1.9 MB (publicsuffix builds a trie). Lookup gap is µs, invisible at FIDO rate. Picked DAFSA-first for the smaller footprint.

libwebauthn/src/ops/webauthn/psl/dafsa.rs
AlfioEmanueleFresta: You're right, changed to BadHeader.

libwebauthn/src/ops/webauthn/psl/dafsa.rs
AlfioEmanueleFresta: Changed this  to the new BadHeader as per your suggestion below! Ty :)

PR #216, opened 2026-05-14
feat(cable): hybrid transport over a direct BLE L2CAP channel

AlfioEmanueleFresta:
Adds support for the CTAP 2.3 hybrid transport over a direct BLE L2CAP channel, with no tunnel server.

Commits 1-2 refactor the Noise handshake and encrypted framing onto a transport-agnostic data-channel abstraction. The rest add advertisement-suffix parsing, the QR key 6 transport list, a `bluer`-backed L2CAP data channel (l2cap feature only, no D-Bus), and L2CAP-or-WebSocket selection in `connection_stage`.

Scope is QR-initiated only. The draft has no way to negotiate a BLE channel for state-assisted transactions, so those stay on WebSocket.

Validated on Android 16 (Pixel 8 with Google Play Services). Checked for regressions on iOS.

PR #217, opened 2026-05-17
fix(cable): propagate post-handshake errors from protocol::connection

AlfioEmanueleFresta:
## Summary

Stacked on `feat/pxp-ble` (#216).

`protocol::connection` previously returned `()` and swallowed every post-handshake fault via `let _ = ...`: Noise decrypt/encrypt failures, malformed framing inside the encrypted channel, dropped CBOR-response receivers and transport-level I/O errors. The connection silently degraded (subsequent frames keep failing the same way) and the channel eventually moved to `Terminated` without ever emitting `CableUxUpdate::CableUpdate(CableUpdate::Error(_))`. Subscribers had no way to distinguish a clean close from a fault.

## Behaviour changes

- `protocol::connection` returns `Result<(), TransportError>`. Each arm of the select loop is explicit about whether a condition is fatal (exit) or benign (continue + log).
- A new `TransportError::EncryptionFailed` variant is emitted for Noise transport-mode encrypt/decrypt failures, distinct from handshake-time `NegotiationFailed` and generic `ConnectionFailed`.
- A peer-sent `CableTunnelMessageType::Shutdown` is treated as a clean close (returns `Ok(())`) instead of an "unexpected" error.
- Malformed or unsigned linking updates are logged and dropped without tearing the channel down: the Noise state is intact and only that one update is unusable. Matches Chromium.
- `qr_code_device` and `known_devices` route `Err(_)` from the tunnel loop through `UxUpdateSender::send_error`, which already transitions to `Terminated`.

## Test plan

- [x] `cargo check --workspace --all-targets`
- [x] `cargo test --workspace --all-targets`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`

### DIFF
--- a/libwebauthn/src/transport/cable/protocol.rs
+++ b/libwebauthn/src/transport/cable/protocol.rs
@@ -117,6 +117,14 @@ enum CableTunnelMessageType {
     Update = 2,
 }
 
+/// Result of processing a single inbound tunnel frame.
+enum RecvOutcome {
+    /// Frame handled, keep the loop running.
+    Continue,
+    /// Peer sent a `Shutdown` control message; close the channel cleanly.
+    PeerShutdown,
+}
+
 #[derive(Clone)]
 pub(crate) enum CableTunnelConnectionType {
     QrCode {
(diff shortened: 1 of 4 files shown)


libwebauthn/src/transport/cable/protocol.rs
```diff
@@ -539,27 +569,24 @@ async fn connection_recv(
...
+) -> Result<RecvOutcome, TransportError> {
     let decrypted_frame = decrypt_frame(encrypted_frame, noise_state).await?;
 
-    // TODO handle the decrypted frame
     let cable_message: CableTunnelMessage = match CableTunnelMessage::from_slice(&decrypted_frame) {
         Ok(cable_message) => cable_message,
         Err(e) => {
             error!(?e, "Failed to decode CABLE tunnel message");
-            return Err(Error::Transport(TransportError::ConnectionFailed));
+            return Err(TransportError::InvalidFraming);
```
msirringhaus: This is a bit convoluted. `CableTunnelMessage::from_slice()` only returns `Error::Transport(TransportError::InvalidFraming)` on error, which we throw away here and hardcode it to `TransportError::InvalidFraming` again. `from_slice()` could just return a `TransportError` which we log here and send along.

libwebauthn/src/transport/cable/protocol.rs
```diff
@@ -628,15 +659,16 @@ fn parse_known_device(
     tunnel_domain: &str,
     linking_info: &CableLinkingInfo,
     noise_state: &TunnelNoiseState,
-) -> Result<CableKnownDeviceInfo, Error> {
-    let known_device = CableKnownDeviceInfo::new(tunnel_domain, linking_info)?;
+) -> Result<CableKnownDeviceInfo, TransportError> {
+    let known_device = CableKnownDeviceInfo::new(tunnel_domain, linking_info)
+        .map_err(|_| TransportError::InvalidFraming)?;
```
msirringhaus: Same here. `CableKnownDeviceInfo::new()` returns `InvalidFraming` anyways, and we throw away the error and hardcode it again.

msirringhaus (changes requested):
If I'm not mistaken, some things could be simplified a bit

PR #218, opened 2026-05-17
fix(cable): preserve I/O error kind on the WebSocket data channel

AlfioEmanueleFresta:
## Summary
- Map `tungstenite::Error::Io` to `TransportError::IoError(kind)` in `WebSocketDataChannel::{send, recv}`, matching `L2capDataChannel`'s variant strategy so the `CableDataChannel` trait abstraction lets callers differentiate failure modes regardless of transport.
- Treat `tungstenite::Error::ConnectionClosed` as a clean close in `recv` (returns `Ok(None)`), matching how `Message::Close` is already handled.
- Protocol-level issues (unexpected message type, `Tls`, `Capacity`, `Protocol`, `Utf8`, `Url`, `Http`, `AlreadyClosed`, etc.) still map to `TransportError::ConnectionFailed`.

Stacked on top of #216.

## Test plan
- [x] `cargo build --workspace --all-features`
- [x] `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- [x] `cargo test --workspace --all-features --lib` (258 passed)

### DIFF
--- a/libwebauthn/src/transport/cable/data_channel.rs
+++ b/libwebauthn/src/transport/cable/data_channel.rs
@@ -2,7 +2,7 @@
 use async_trait::async_trait;
 use futures::{SinkExt, StreamExt};
 use tokio::net::TcpStream;
-use tokio_tungstenite::tungstenite::Message;
+use tokio_tungstenite::tungstenite::{Error, Message};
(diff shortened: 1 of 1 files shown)


msirringhaus (approved):
lgtm

PR #219, opened 2026-05-17
feat(webauthn): related-origins validation (WebAuthn L3 §5.11)

### LINKED ISSUE #160: Implement Related Origins support (WebAuthn L3 § 5.11)
## Summary

WebAuthn Level 3 introduces [Related Origins](https://www.w3.org/TR/webauthn-3/#sctn-related-origins) (§ 5.11), allowing credentials to be used across unrelated domains. The RP hosts a JSON document at `https://{rpId}/.well-known/webauthn` listing allowed origins.

## Proposed Implementation

1. **New module**: `ops/webauthn/related_origins.rs`
   - `RelatedOriginsDocument` struct for parsing the well-known JSON
   - `RelatedOriginsHttpClient` trait for fetching (pluggable, testable)
   - `validate_related_origins()` implementing the validation procedure from § 5.11.1

2. **Enhance `RelyingPartyId`** (addresses #137)
   - Add `validate_for_origin()` to check if RP ID is a registrable suffix
   - Return whether related origins validation is required

3. **Dependencies**
   - `url` for origin parsing
   - `publicsuffix` for registrable domain detection

## Feature Flags

```toml
[features]
related-origins = []
related-origins-client = ["related-origins", "reqwest"]
```

- `related-origins`: Core validation logic with trait-based HTTP client
- `related-origins-client`: Optional embedded HTTP client for testing/convenience

## Future Consideration

The existing WebSocket/Noise implementation for caBLE could be refactored to follow the same pattern (trait + optional default implementation behind a feature flag) for consistency.

AlfioEmanueleFresta:
Implements WebAuthn L3 §5.11 "Related Origins". When a request's rp.id is not a registrable suffix of the caller's effective domain, libwebauthn resolves the relying party's allowed origins and accepts the request if one matches the caller.

Origin resolution is pluggable and the matching always runs in libwebauthn. An optional reqwest-backed source fetches the RP's `.well-known/webauthn` document behind the `reqwest-related-origins-source` feature, so the core crate stays HTTP-client-free. A caller that already has the list, such as a browser, can supply its own source and skip the fetch. Related origins can also be turned off.

Request building now takes its public-suffix-list and related-origins dependencies through a settings value rather than positional parameters, which keeps the API manageable as options grow. Existing call sites need a small update for the new signature.

Closes #160. Based on #173 by @HarveyOrourke15.


libwebauthn/src/ops/webauthn/related_origins/mod.rs
```diff
@@ -0,0 +1,744 @@
...
+/// WebAuthn L3 §5.11 minimum; capped at 5 to bound abuse surface.
+const MAX_REGISTRABLE_LABELS: usize = 5;
+
+#[derive(Debug, Clone)]
+pub struct WellKnownResponse {
+    pub content_type: Option<String>,
+    pub body: Vec<u8>,
+}
+
+/// Fetcher for `https://{rp_id}/.well-known/webauthn`, per WebAuthn L3 §5.11.1
```
msirringhaus: Not sure how relevant this is, as we don't expect to get a lot of implementations for this, other than reqwest, but:
I wonder if it would be 'safer' to have the client JUST do the http-request and nothing else.
Meaning: `validate_related_origins()` will already create the well-known-url, `fetch_well_known()` (which would then just be `fetch()`) would simply fetch the website (not even knowing what URL that is), return status code, transport-type and optionally content back to `validate_related_origins()`, which does *all* the checks.
Doing that would make it way harder to mess up an implementation (e.g. forgetting to implement some checks that are only documented in the comment here). The only contract would then be to follow redirects and cap body size.

msirringhaus (approved):
Looking good overall. One architectural question, but I'm fine with the way it is currently, too.

libwebauthn/src/ops/webauthn/related_origins/mod.rs
AlfioEmanueleFresta: Thank you. Adopted this. The fetcher is now a minimal HttpClient: it does the GET and hands back the status, headers, and body, nothing else. It has no idea about the well-known semantics. 

The content-type and JSON checks and the origin matching all moved out of it, so an implementation only has to do the request, follow https-only redirects, and cap the body.

PR #224, opened 2026-05-19
feat(webauthn): largeBlob write and delete

AlfioEmanueleFresta:
## Summary

Adds the WebAuthn L3 §10.1.5 `largeBlob.write` extension and a libwebauthn-side delete operation, layered on top of CTAP 2.2 §6.10.6. The platform fetches the existing on-authenticator array, replaces or erases the entry that decrypts under the credential's `largeBlobKey`, and re-uploads the canonical CBOR array via chunked `authenticatorLargeBlobs(set)`. Foreign entries (other credentials' blobs, unknown CBOR fields) are preserved byte-for-byte per the §6.10.2 platform contract.

Stacked on #206 (read leg). Merge that first.

## Scope

- New `Write(Vec<u8>)` and `Delete` variants on the get-assertion largeBlob extension input
- `written` flag on the assertion's largeBlob output
- `pinUvAuthParam` for each chunk derived per §6.10.2: `authenticate(token, 32×0xff || h'0c00' || u32_le(offset) || SHA-256(set))`
- `lbw` permission negotiated into the get-assertion token when write or delete is requested, so one PIN/UV ceremony covers both operations. Unprotected authenticators continue to accept the write without auth params per §6.10.2 line 137
- L3 mutual-exclusion enforced: `read`+`write` together returns a NotSupportedError, and write/delete require exactly one `allowCredentials` entry
- Delete with no matching entry reports `written=false`, matching the strict §6.10.6 "return an error" branch

## Tests

- Unit tests cover the `pinUvAuthParam` byte construction, AAD format, the read-modify-write loop, and the byte-fingerprint of the canonical empty array (`h'8076be8b528d0075f7aae98d6fa57a6d3c'`)
- Integration tests against the virt authenticator: write then read round trip, write replaces existing, delete erases, delete with no prior entry reports a failure flag

## Follow-ups

- Optional W3C webauthn#2337-style UV auto-upgrade for write/delete on PIN-protected devices under UV=Discouraged. Currently the platform falls through to `written=false` rather than forcing a PIN prompt
- §6.10.7 garbage collection of orphan blobs after credential deletion is not implemented

### DIFF
--- a/libwebauthn-tests/tests/large_blob.rs
+++ b/libwebauthn-tests/tests/large_blob.rs
@@ -12,7 +12,7 @@ use libwebauthn::proto::ctap2::{
     Ctap2PublicKeyCredentialDescriptor, Ctap2PublicKeyCredentialRpEntity,
     Ctap2PublicKeyCredentialUserEntity,
 };
-use libwebauthn::transport::{Channel, Device};
+use libwebauthn::transport::{Channel, ChannelSettings, Device};
 use libwebauthn::webauthn::WebAuthn;
 use libwebauthn::UvUpdate;
 use libwebauthn_tests::virt::get_virtual_device;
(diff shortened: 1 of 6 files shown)


AlfioEmanueleFresta:
Auto-closed: a mis-targeted rebase briefly pushed this branch to the same commit as its base, so GitHub closed it for having no diff. The branch content is intact and correct. This PR cannot be reopened (its head and base are frozen equal), so it is superseded by #261, which has the same branch and the up-to-date content.

PR #228, opened 2026-05-26
fix(ctap2): pass through arbitrary COSE algorithm identifiers

### LINKED ISSUE #148: Passthrough unknown key algorithms

AlfioEmanueleFresta:
The COSE algorithm identifier type was a closed enum with a sentinel for unknown values. Any algorithm a relying party requested that wasn't one of three known values got silently rewritten to the sentinel before reaching the authenticator. Authenticators that supported RS256, ES384, PS256, secp256k1 or other algorithms never saw the requested algorithm.

The type now passes arbitrary algorithm identifiers through unchanged, with named constants for the IANA-registered values libwebauthn cares about. The `-9` codepoint was historically misnamed for an internal use and is now aligned with RFC 9864 ESP256, the actual IANA assignment.

This is a breaking change on the public Rust API.

Closes #148.

### DIFF
--- a/libwebauthn/src/proto/ctap2/model.rs
+++ b/libwebauthn/src/proto/ctap2/model.rs
@@ -7,7 +7,7 @@ use crate::{
 use num_enum::{IntoPrimitive, TryFromPrimitive};
 use serde_bytes::ByteBuf;
 use serde_derive::{Deserialize, Serialize};
-use serde_repr::{Deserialize_repr, Serialize_repr};
+use serde_repr::Serialize_repr;
 
 mod get_info;
 pub use get_info::Ctap2GetInfoResponse;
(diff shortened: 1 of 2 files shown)


libwebauthn/src/proto/ctap2/model.rs
```diff
@@ -172,14 +172,92 @@ pub struct Ctap2PublicKeyCredentialDescriptor {
...
+            Self::PS512 => Some("PS512"),
+            Self::ES256K => Some("ES256K"),
+            Self::RS256 => Some("RS256"),
+            Self::RS384 => Some("RS384"),
+            Self::RS512 => Some("RS512"),
+            Self::RS1 => Some("RS1"),
+            _ => None,
+        };
+        match name {
+            Some(n) => write!(f, "Ctap2COSEAlgorithmIdentifier::{}", n),
```
msirringhaus: For debugging, it would probably be nice to have both the name and the value in there.

msirringhaus (approved):
lgtm!

AlfioEmanueleFresta:
Thanks! Addressed your comment.

PR #229, opened 2026-05-26
fix(ctap2): store credential public keys as opaque COSE bytes

AlfioEmanueleFresta:
The credential public key returned by an authenticator was being stored in a closed Rust enum that only handled P-256 and Ed25519. Any credential using RSA, P-384, secp256k1 or a post-quantum algorithm failed to deserialise, so the whole registration response was rejected.

Credential public keys are now stored as raw COSE bytes. The platform doesn't crypto-validate them, it forwards them to the relying party, so the closed enum was over-modelling. The authenticator data parser captures the COSE_Key bytes verbatim, preserving the attestation signature's bit-identity for relying-party verification.

A small helper reads the mandatory `alg` parameter per WebAuthn L3 §6.5.2, which is what populates `publicKeyAlgorithm` in the JSON response. The previous ES256 fallback when attested credential data was missing is removed because the AT flag MUST be set per CTAP §6.1, and a silent fallback was hiding real protocol violations.

The closed enum is still used for the ECDH-ES P-256 key agreement during PIN/UV protocols, where the COSE shape is fixed by spec.

Stacked on top of #228.

### DIFF
--- a/libwebauthn/src/ops/webauthn/make_credential.rs
+++ b/libwebauthn/src/ops/webauthn/make_credential.rs
@@ -26,8 +26,8 @@ use crate::{
     proto::{
         ctap1::{Ctap1RegisteredKey, Ctap1Version},
         ctap2::{
-            cbor, Ctap2AttestationStatement, Ctap2COSEAlgorithmIdentifier, Ctap2CredentialType,
-            Ctap2GetInfoResponse, Ctap2MakeCredentialsResponseExtensions,
+            cbor, cose, Ctap2AttestationStatement, Ctap2COSEAlgorithmIdentifier,
+            Ctap2CredentialType, Ctap2GetInfoResponse, Ctap2MakeCredentialsResponseExtensions,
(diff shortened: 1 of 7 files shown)


libwebauthn/src/ops/webauthn/make_credential.rs
```diff
@@ -67,42 +67,33 @@ impl WebAuthnIDLResponse for MakeCredentialResponse {
         &self,
         request: &Self::Context,
     ) -> Result<Self::IdlModel, ResponseSerializationError> {
-        // Get credential ID from attested credential data
-        let credential_id_bytes = self
+        // The AT flag MUST be set on makeCredential responses per CTAP §6.1.
```
msirringhaus: Either provide a link or specify which CTAP-version is meant here, because §6 is something different for each release. A direct quote would probably also be nice, because I couldn't find it right away.

msirringhaus (approved):
lgtm. One comment update would be nice, though.

PR #230, opened 2026-05-26
feat(webauthn): emit SubjectPublicKeyInfo from getPublicKey()

AlfioEmanueleFresta:
WebAuthn L3 §5.2.1.1 says `AuthenticatorAttestationResponse.getPublicKey()` returns DER-encoded SubjectPublicKeyInfo for credentials using ES256, EdDSA Ed25519 and RS256, and null for any algorithm the user agent does not implement. libwebauthn was emitting raw COSE bytes there instead, which relying parties could not feed into `SubtleCrypto` or standard X.509 parsers.

A new converter produces SPKI for the WebAuthn L3 floor plus ESP256, which is equivalent to ES256 per RFC 9864. Algorithms outside the supported set return null per spec. Malformed keys for understood algorithms surface as an error.

The conversion uses the `spki` and `der` crates from RustCrypto, both already in the transitive dependency tree via `p256`, so no new external runtime crates land. Adding new algorithm support later is a small per-algorithm addition.

Stacked on top of #229.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -839,7 +839,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "79b71cca7d95d7681a4b3b9cdf63c8dbc3730d0584c2c74e31416d64a90493f4"
 dependencies = [
  "crypto-bigint 0.2.5",
- "der_derive",
+ "der_derive 0.4.1",
 ]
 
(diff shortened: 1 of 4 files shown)


msirringhaus (approved):
I'm not a cryptography expert, but lgtm

PR #264, opened 2026-06-09
fix(credmgmt): Yubikey error when NextCred is not handled statefully

Nisker:
I have a Token2 and a Yubikey "Security Key C NFC".
The Yubikey is more strict than the Token2 and causes errors.

 `cargo run --example cred_management_hid` with option 2:  "(2) Enumerate credentials"
causes `Error: Ctap(NotAllowed)` on the Yubikey after the PIN have been input.
I think this is due to enumerateCredentialsGetNextCredential and enumerateRPsGetNextRP being statefull in the fido v2.1 spec.

### DIFF
--- a/libwebauthn/src/management/credential_management.rs
+++ b/libwebauthn/src/management/credential_management.rs
@@ -121,25 +121,8 @@ where
     }
 
     async fn enumerate_rps_next_rp(&mut self, timeout: Duration) -> Result<Ctap2RPData, Error> {
-        let mut req = Ctap2CredentialManagementRequest::new_enumerate_rps_next_rp();
-        let resp = loop {
-            let uv_auth_used = user_verification(
-                self,
-                UserVerificationRequirement::Preferred,
-                &mut req,
-                timeout,
-            )
-            .await?;
-
-            // On success, this is an all-empty Ctap2AuthenticatorConfigResponse
-            handle_errors!(
-                self,
-                self.ctap2_credential_management(&req, timeout).await,
-                uv_auth_used,
-                timeout,
-                req
-            )
-        }?;
+        let req = Ctap2CredentialManagementRequest::new_enumerate_rps_next_rp();
+        let resp = self.ctap2_credential_management(&req, timeout).await?;
         Ok(Ctap2RPData::new(
             unwrap_field!(resp.rp),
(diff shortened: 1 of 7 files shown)


AlfioEmanueleFresta (approved):
Thanks for the fix and the clear diagnosis @Nisker, you read the spec exactly right.

- CTAP 2.1 §6.8.3 and §6.8.4 send the GetNext subcommands with only the subCommand parameter.
- The stateful commands section lets authenticators fail a continuation with CTAP2_ERR_NOT_ALLOWED unless it directly follows Begin or another GetNext.
- Our old loop broke this in two ways. It attached a pinUvAuthParam, and it sent a getInfo between Begin and GetNext.
- YubiKeys enforce the strict reading, the Token2 is lenient.
- libfido2, Chromium and python-fido2 all send GetNext bare.

Reproduced here on a YubiKey with two passkeys under one RP. On master:

```
DEBUG user_verification:ctap2_get_info: CTAP2 GetInfo successful
TRACE ctap2_credential_management: request=Ctap2CredentialManagementRequest {
  subcommand: Some(EnumerateCredentialsGetNextCredential), subcommand_params: None,
  protocol: Some(Two), uv_auth_param: Some([155, 189, 175, ..]) }
DEBUG ctap2_credential_management: Received CBOR response status=NotAllowed
Error: Ctap(NotAllowed)
```

With this PR, same device and credentials:

```
TRACE ctap2_credential_management: request=Ctap2CredentialManagementRequest {
  subcommand: Some(EnumerateCredentialsGetNextCredential), subcommand_params: None,
  protocol: None, uv_auth_param: None }
DEBUG ctap2_credential_management: Received CBOR response status=Ok
DEBUG ctap2_credential_management: CTAP2 CredentialManagement successful
```

I pushed two commits on top of your branch:

- Regression tests pinning the wire format. They fail on master.
- A fix for authenticators that only advertise credentialMgmtPreview. Begin resolves the 0x41 preview command from getInfo, but GetNext no longer went through that path, so it would have gone out as 0x0A. Begin now records the command set it resolved and GetNext reuses it.

Nice catch, thanks again!

AlfioEmanueleFresta:
This is now released in libwebauthn 0.7.1 if you want to test, thanks again @Nisker!

PR #265, opened 2026-06-09
fix(ctap2): do not send uv option together with pinUvAuthParam

Nisker:
This was tested with libwebauthn 0.5.1 in credentialsd master branch ([7c13749](https://github.com/linux-credentials/credentialsd/commit/7c137495b0fe1651171cc21a830feb0bfb7ff69c)) in a modified Teams for linux (FIDO2 passkey login login).
Token2 version 3.1 logins without problems.
Yubikey triggers `Ctap(InvalidOption)` error right after a PIN has been entered.

The FIDO 2.1 spec.`§6.1.2. authenticatorMakeCredential` and `§6.2.2. authenticatorGetAssertion` says "If the [pinUvAuthParam](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#getassert-pinuvauthparam) is present, let the "[uv](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#getassert-uv)" [option](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#getassert-option-key) be treated as being present with the value false."
So having UV=true and pinUvAuthParam will cause errors on a strict key like Yubikey.

### DIFF
--- a/libwebauthn/src/proto/ctap2/model/get_assertion.rs
+++ b/libwebauthn/src/proto/ctap2/model/get_assertion.rs
@@ -692,6 +692,37 @@ mod tests {
     }
 
     #[test]
+    fn pin_uv_auth_param_clears_uv_option() {
+        use crate::ops::webauthn::UserVerificationRequirement;
+        use crate::pin::PinUvAuthProtocolOne;
+
+        let mut request = make_request(vec![]);
+        request.user_verification = UserVerificationRequirement::Required;
+        let mut ctap2 = Ctap2GetAssertionRequest::from(request);
+        assert!(ctap2.options.unwrap().require_user_verification);
+
+        let proto = PinUvAuthProtocolOne::new();
+        ctap2
+            .calculate_and_set_uv_auth(&proto, &[0xAA; 32])
+            .unwrap();
+
+        assert!(ctap2.pin_auth_param.is_some());
+        assert!(!ctap2.options.unwrap().require_user_verification);
+
+        // Wire check: the options map (0x05) must not contain "uv".
+        let bytes = crate::proto::ctap2::cbor::to_vec(&ctap2).unwrap();
+        let parsed: BTreeMap<u64, Value> = crate::proto::ctap2::cbor::from_slice(&bytes).unwrap();
+        let Some(Value::Map(options)) = parsed.get(&0x05) else {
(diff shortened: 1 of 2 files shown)


AlfioEmanueleFresta (approved):
Thanks for the fix and the clear report @Nisker, and for testing against both Token2 and YubiKey hardware. This is spec-correct. The CTAP 2.1 options tables for authenticatorMakeCredential and authenticatorGetAssertion both state: "Platforms MUST NOT include both the 'uv' option key and the pinUvAuthParam parameter in the same request." 

YubiKeys seem to enforce this strictly, which is exactly the `CTAP2_ERR_INVALID_OPTION` you saw. Reproduced here on a YubiKey (firmware 5.x, `userVerification: "required"` assertion after PIN entry). On master:

```
TRACE ctap2_get_assertion: request=Ctap2GetAssertionRequest { relying_party_id: "example.org", [..]
  options: Some(Ctap2GetAssertionOptions { require_user_presence: true, require_user_verification: true }),
  pin_auth_param: Some([184, 209, 64, ..]), pin_auth_proto: Some(2) }
DEBUG ctap2_get_assertion: Received CBOR response status=InvalidOption
```

With this PR, same device and request:

```
TRACE ctap2_get_assertion: request=Ctap2GetAssertionRequest { relying_party_id: "example.org", [..]
  options: Some(Ctap2GetAssertionOptions { require_user_presence: true, require_user_verification: false }),
  pin_auth_param: Some([208, 153, 182, ..]), pin_auth_proto: Some(2) }
DEBUG ctap2_get_assertion: CTAP2 GetAssertion successful
  flags: AuthenticatorDataFlags(USER_PRESENT | USER_VERIFIED)
```

User verification still happens through the pinUvAuthToken, as the UV flag in the response shows.

I pushed a small commit on top of your branch adding regression tests for both request types, and tweaked the PR title to match our commit conventions. Nice catch, thanks again!

AlfioEmanueleFresta:
This is now released in libwebauthn 0.7.1 if you want to test, thanks again @Nisker!

Issue #182, opened 2026-04-24
How to deal with authenticators with multiple CTAP transports on one physical interface


iinuwa:
Somewhat related to #154 is that if a device supports the CCID/smart-card protocol over USB as well as HID, then it will appear as two devices without a way to deduplicate them[^1].

[^1]: I think this needs CTAP spec changes to make this robust (e.g., returning a random ID from getInfo on every power cycle to identify the same device across transports).

But I think that we can improve this situation by only using physical interface-specific APIs. USB and BLE are fine: we should use libnfc only for the `nfc::is_available()`, or return `false`. (The reason for this is that PC/SC abstracts over physical transports, so we can't know at that level whether it's NFC, USB or a smartcard on some other port.) 

- We can still support reading NFC keys with just pcsc, we just won't be able to tell inform the UI that they should say "a USB or NFC security key" rather than just "a USB security key". Similarly, this allows us to use "smart-card only" authenticators, though I don't know if any of those exist.

(I think this issue motivates a change in credentialsd though: instead of selecting transports for security keys, we should just generically show "security key" and preemptively discovery on all available transports. That'll be a separate issue in credentialsd.)

AlfioEmanueleFresta:
Hey @iinuwa, I went digging on this. You're right that the root cause is PC/SC hiding the physical interface, so a USB key's CCID interface looks exactly like an NFC reader.

I'd be a bit wary of leaning on libnfc for the "is NFC available" signal though. libnfc only knows PN53x readers, so it would miss the common modern contactless readers (ACR1252U, OMNIKEY 5022, and friends) and report NFC as unavailable even with a real reader attached. It is also a second, fairly stale dependency that competes with pcscd for exclusive access to the reader. PC/SC, on the other hand, is what the rest of the FIDO ecosystem uses for NFC (libfido2, python-fido2), it covers the broad set of CCID readers, and it is already present on a normal desktop. So my vote is to keep pcsc and not depend on libnfc.

We also do not need to tell NFC and USB apart to fix the duplicate. The two interfaces of one physical key share a USB address, so we can just recognize the same device and drop the duplicate from the NFC list. PC/SC hands us the USB bus and address through `SCARD_ATTR_CHANNEL_ID`, and the HID side resolves the same pair via sysfs. A standalone reader has no HID twin, so it is left alone. This is also the dedup that #154 left open.

```rust
pub struct UsbDeviceId { pub bus: u8, pub address: u8 }

[...]

PR #270, opened 2026-06-15
feat(nfc): dedupe USB-HID authenticators

AlfioEmanueleFresta:
Some security keys appear twice when we go looking for authenticators. The same key shows up once as a USB device over HID, and again as a PC/SC reader for its built-in smart-card interface. Both entries are the same physical key, but until now nothing connected them, so a single key plugged into USB also looked like a separate NFC authenticator waiting to be used.

We now recognize the two entries as one device by comparing the USB address behind each. When a reader resolves to the same physical USB device as a key we already see over HID, we drop it from the NFC list. A key plugged in over USB is no longer offered as a separate NFC device.

Reading real NFC keys is unchanged. A key held to a contactless reader has no matching USB device, so it stays in the list and is read exactly as before.

Relates to #182


iinuwa (approved):
I haven't been able to run this yet, but it looks good to me from reading the source.

PR #282, opened 2026-06-15
feat(transport): advertise NFC and hybrid transports

### LINKED ISSUE #258: feat(transport): advertise NFC/hybrid and report registration transports
## What's wrong

libwebauthn implements four transports (USB/HID, BLE, NFC, caBLE/hybrid) but the public surface advertises only two, and registration responses never report the transport used.

- `Transport` and `available_transports()` in `libwebauthn/src/lib.rs` expose only `Usb` and `Ble`. NFC (`transport::nfc`, gated behind `nfc-backend-pcsc`/`nfc-backend-libnfc`) and caBLE/hybrid (`transport::cable`) are implemented but omitted.
- `MakeCredentialResponse::to_idl_model()` (`ops/webauthn/make_credential.rs`, ~line 104) always sets `transports = Vec::new()`, so `AuthenticatorAttestationResponseJSON.transports` (`ops/webauthn/idl/response.rs`) is always empty.

Note: the public `crate::Transport` enum is distinct from the internal `transport::Transport` marker trait (`transport/transport.rs`), which already covers HID, BLE, `Cable`, and `Nfc`.

## Why it matters

Callers reading `available_transports()` can't discover NFC or hybrid. Per WebAuthn L3 5.2.1.1, RPs persist the response `transports` and replay them as `allowCredentials[].transports` hints, so an empty list degrades later authentication UX. The list must be unique and lexicographically sorted, holding only `AuthenticatorTransport` values (L3 5.8.4: `usb`, `nfc`, `ble`, `hybrid`, `internal`, `smart-card`).

## What to do

- [ ] Add NFC and hybrid/caBLE to the public `Transport` enum (`lib.rs`).
- [ ] Update `available_transports()` to report compiled-in transports. NFC only appears under an `nfc-backend-*` feature. caBLE is always compiled (runtime depends on a BLE adapter, `cable::is_available()`), so decide whether to report compiled-in support or consult `is_available()`.
- [ ] Expose the active transport from `Channel` (`transport/channel.rs`) and pass it into `to_idl_model` for the registration response conversion.
- [ ] Populate `AuthenticatorAttestationResponseJSON.transports` from the active transport, mapping via `Ctap2Transport` (`proto/ctap2/model.rs`), which already serializes the correct lowercase tokens.
- [ ] Ensure the emitted list is deduplicated, lexicographically sorted, and free of unknown values (L3 5.2.1.1). Leave it empty when the transport is unknown.
- [ ] Cover both the FIDO2 and U2F-downgrade registration paths.
- [ ] Update `test_response_to_idl_model` in `make_credential.rs`, which asserts an empty list.

AlfioEmanueleFresta:
The library implements four transports but advertised only USB and BLE, and registration responses always returned an empty transports list. This adds NFC and hybrid to the public transport list, with NFC shown only when a backend is compiled in, and fills the registration transports from the transport actually used, ordered and deduplicated per the WebAuthn rules.

Closes #258.

### DIFF
--- a/libwebauthn/examples/ceremony/webauthn_ble.rs
+++ b/libwebauthn/examples/ceremony/webauthn_ble.rs
@@ -75,11 +75,7 @@ pub async fn main() -> Result<(), Box<dyn Error>> {
                 .unwrap();
         println!("WebAuthn MakeCredential response: {:?}", response);
 
-        match response.to_json_string(
-            &make_credentials_request,
-            channel.transport(),
-            JsonFormat::Prettified,
-        ) {
+        match response.to_json_string(&make_credentials_request, JsonFormat::Prettified) {
             Ok(response_json) => {
                 println!(
                     "WebAuthn MakeCredential response (JSON):\n{}",
@@ -117,11 +113,7 @@ pub async fn main() -> Result<(), Box<dyn Error>> {
         println!("WebAuthn GetAssertion response: {:?}", response);
 
         for assertion in &response.assertions {
-            match assertion.to_json_string(
-                &get_assertion,
-                channel.transport(),
-                JsonFormat::Prettified,
-            ) {
+            match assertion.to_json_string(&get_assertion, JsonFormat::Prettified) {
                 Ok(assertion_json) => {
                     println!("WebAuthn GetAssertion response (JSON):\n{}", assertion_json);
                 }
--- a/libwebauthn/examples/ceremony/webauthn_ble.rs
+++ b/libwebauthn/examples/ceremony/webauthn_ble.rs
@@ -75,7 +75,11 @@ pub async fn main() -> Result<(), Box<dyn Error>> {
                 .unwrap();
         println!("WebAuthn MakeCredential response: {:?}", response);
 
-        match response.to_json_string(&make_credentials_request, JsonFormat::Prettified) {
+        match response.to_json_string(
+            &make_credentials_request,
+            channel.transport(),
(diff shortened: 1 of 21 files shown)


iinuwa:
@AlfioEmanueleFresta, my read of the WebAuthn spec says that `transports` should be all the transports supported by the authenticator. Since libwebauthn doesn't expose authenticator get info directly, does it make sense to have the transports be returned by libwebauthn (e.g. in the MakeCredentialResponse) instead of the caller specifying it?

AlfioEmanueleFresta:
Thanks @iinuwa, agreed. I switched it to do exactly that. The registration `transports` now come from the authenticator's getInfo (member 0x09, which libwebauthn already decodes), populated on `MakeCredentialResponse` rather than specified by the caller, and folded together with the transport the ceremony actually ran over. The list is emitted as unique tokens in lexicographical order per WebAuthn L3 §5.2.1.1.

This follows the spec's `getTransports()` definition (the transport used to register, unioned with the transports the client believes the authenticator supports) and matches Chromium, which folds getInfo 0x09 into that set. For authenticators that don't report 0x09 (CTAP 2.0) or on the U2F downgrade path, the union reduces to just the ceremony transport, which preserves the prior behavior.

iinuwa:
Yeah, that sounds good to union those, though it would be weird if an authenticator didn't specify that it supports the transport it just used.

iinuwa (approved):
LGTM!

PR #304, opened 2026-07-22
chore(nfc): Bump nfc1 to 0.7.1

iinuwa:
This was previously pinned to 0.6.0 because of build issues for non-vendored build in 0.6.1. 0.7.1 has been release which fixes that issue, so we're removing the pin to let consumers resolve their own version.


michaelbeaumont:
@iinuwa anything holding back this PR?

PR #306, opened 2026-08-07
Remove duplicate license specification

abergmeier:
Specifying both license and license-file is invalid. license-file seems the more concise option of the two.


libwebauthn/Cargo.toml
```diff
@@ -9,7 +9,6 @@ authors = [
 ]
 edition = "2021"
 rust-version = "1.88"
-license = "LGPL-2.1-or-later"
```
michaelbeaumont, line 12: this is a valid SPDX identifier though, why the license file exactly?

PR #308, opened 2026-09-01
Do not print how many devices where found for each loop iteration

msirringhaus:
See https://github.com/linux-credentials/credentialsd/issues/221


AlfioEmanueleFresta:
I have a slight preference for reducing this to debug, so it's disabled by default but can be still enabled for this crate easily (it should not interfere with credentialsd debug logging).

Happy for this to be merged either way!

AlfioEmanueleFresta (approved):

msirringhaus:
Would you keep both debug-logs then? Or reduce the one to debug and the other one to trace?

AlfioEmanueleFresta:
Debug and trace sound ideal.

Thank you!

iinuwa (approved):
Works for me!

